PR #2, opened 2015-07-15
Gcw 442 browse goods top level category 

swatijadhav:


ember-cli-build.js
```diff
@@ -29,5 +29,11 @@ module.exports = function(defaults) {
   // please specify an object with the list of modules as keys
   // along with the exports of each module as its value.
 
+  app.import('bower_components/blueimp-file-upload/js/vendor/jquery.ui.widget.js');
+  app.import('bower_components/blueimp-file-upload/js/jquery.iframe-transport.js');
+  app.import('bower_components/cloudinary/js/load-image.min.js');
+  app.import('bower_components/cloudinary/js/canvas-to-blob.min.js');
+  app.import('bower_components/cloudinary/js/jquery.cloudinary.js');
+
```
mcm-ham: For displaying images (no upload) I think we only need to import jquery.cloudinary.js.

swatijadhav: yeah @mcm-ham 
Also we want to display some transformations to images, like on browse page, we will be showing the main categories images in circular format. For those transformation we will need other js right? (am not much sure about which js files it would need exactly)

mcm-ham: @swati I see, in that case I think we only need `canvas-to-blob.js`. The blueimp-file-upload will be to do with file uploading of course, and apparently [load-image.js](https://github.com/blueimp/JavaScript-Load-Image) allows you to display images from file upload controls, so don't think we need it.

PR #14, opened 2015-08-06
GCW-539 Comment# Added i18n initializer to access ember-i8n across al…

shivanibhanwal:
Added i18n initialiser to access ember-i8n across al files in ember
removed all hardcoded strings and added content to translation files.


steveyken:
Thanks Shivani, looks good to me!

swatijadhav:
:+1:

PR #19, opened 2016-03-21
Upgraded browse app to

shivanibhanwal:
Ember 2.4.2
Ember-cli 2.4.1
Ember-data 2.4.0
Ember-cli-qunit 1.2.1
Updated all the test cases as per the new changes in calling
, new syntax for sequence etc
Made change in factory replaced  to


swatijadhav:
:+1:

PR #107, opened 2017-10-03
Package model computed properties test cases

namrataukirde:


tests/unit/models/package-category-test.js
```diff
@@ -51,3 +51,23 @@ test('Check attributes', function(assert){
...
 });
+
+test('nameItemsCount: Returns name and items count', function(assert){
+  var packageCategory,store;
+
+  store = this.store();
+
+  Ember.run(function(){
+    packageCategory = store.createRecord('packageCategory', { id: 1, name: 'Name'});
+  });
```
trojanh: since we are using this.store() only once, can we avoid `store` variable?

namrataukirde: 👍

PR #174, opened 2018-06-27
Edited translation from line 1-116

bryanfzcai:


steveyken:
The merge check that failed was related to duplicate code in codeclimate. not something to fix on translations.js

PR #203, opened 2018-09-19
GCW-2131 New browse homepage

patrixr:
Homepage based on the new design.

Notes:
- Buttons linking to "Book appointment" are currently greyed out with a 'coming soon' message
- Some missing pages were added but are not yet filled in : FAQ, About, Privacy, Terms


Darshan-Chauhan:
Hi @patrixr , I've only seen some of the code, haven't reviewed full PR yet. 
Couple of things I noticed are
- You've used 2 spaces/ 4 spaces for tab. We use 2 space for all our projects.
- Also I think `banner-logo` component is misuse of component concept. Components are generally created when we need to add some custom/additional functionality. For `banner-logo` I think it would be better to use partial as it only contains design and not any custom functionality. 

Overall code looks good. 👍 

CC: @namrataukirde , @abulsayyad123

patrixr:
> Hi @patrixr , I've only seen some of the code, haven't reviewed full PR yet.
> Couple of things I noticed are
> 
> * You've used 2 spaces/ 4 spaces for tab. We use 2 space for all our projects.
> * Also I think `logo-banner` component is misuse of component concept. Components are generally created when we need to add some custom/additional functionality. For `logo-banner` I think it would be better to use partial as it only contains design and not any custom functionality.
> 
> Overall code looks good. 👍
> 
> CC: @namrataukirde , @abulsayyad123

@Darshan-Chauhan Regarding the component, the point was to be able to pass an option/class to it. Could that be done with a partial ?

Darshan-Chauhan:
@patrixr , no you cant pass options to partial. All the options/properties available in hbs(from which you're rendering partial) file will be available in that partial. So you can set option in controller which will be available in partial

patrixr:
> @patrixr , no you cant pass options to partial. All the options/properties available in hbs(from which you're rendering partial) file will be available in that partial. So you can set option in controller which will be available in partial

That wouldn't be very useful in this case, since the logo appears twice in the hbs, one for mobile, and one for desktop, if the value is in the controller, both of them would fetch the same value and render the same way.
I'm not sure I agree with this, as it feels like I'm jumping through hoops just to avoid placing the .hbs file in the component folder, but I'll make an alternative using css classes until I get a better understanding of Ember 😄

Darshan-Chauhan:
@patrixr , rather than adding some patch fix let it be a component. Also for this use case use of component looks better than using partial.
👍

app/styles/templates/_home.scss
```diff
@@ -0,0 +1,106 @@
+.home_page {
+
+    .scroller {
+        height: 93vh;
```
namrataukirde: Could you please use 2 spaces indentation instead of 4 spaces?

PR #207, opened 2018-09-26
GCW-2132-New User Signup feature added

abulsayyad123:
Hi Team,
This is PR for New User Sign up feature. https://jira.crossroads.org.hk/browse/GCW-2132.
And some other changes added which includes.
- User account details page added.
- Search GC organisation page added for account details page.
- User redirection to browse and account-details page according to completion of user details.
- Test cases added.
- New styling added according to new mockup design.

Please review this PR and let me know if any change is required.

Thanks

### DIFF
--- a/app/controllers/account_details.js
+++ b/app/controllers/account_details.js
@@ -9,6 +9,20 @@ export default Ember.Controller.extend({
   messageBox: Ember.inject.service(),
   organisationId: Ember.computed.alias('model.organisation.id'),
   organisationsUserId: Ember.computed.alias('model.organisationsUser.id'),
+  userTitle: Ember.computed.alias('model.user.title'),
+
+  selectedTitle: Ember.computed('userTitle', function (){
+    return{ name: this.get('userTitle'), id: this.get('userTitle')};
+  }),
+
+  titles: Ember.computed(function(){
+    return [
+      { name: "Mr", id: "Mr" },
+      { name: "Mrs", id: "Mrs" },
+      { name: "Miss", id: "Miss" },
+      { name: "Ms", id: "Ms" }
+    ];
+  }),
 
   clearFormData() {
     this.set('organisationName', "");
@@ -39,7 +53,7 @@ export default Ember.Controller.extend({
       var organisationId = this.get('organisationId');
       var position = this.get('model.organisationsUser.position');
       var email = this.get('model.user.email');
-      var title = this.get('model.user.title');
+      var title = this.get('selectedTitle.name');
       new AjaxPromise("/organisations_users", "POST", this.get('session.authToken'), { organisations_user: {
         organisation_id: organisationId, position: position, user_attributes: { first_name: firstName,
         last_name: lastName, mobile: mobilePhone, email: email, title: title }}}).then(data =>{
--- a/app/controllers/account_details.js
+++ b/app/controllers/account_details.js
@@ -1,9 +1,18 @@
 import Ember from "ember";
 
 export default Ember.Controller.extend({
+  authenticate:Ember.inject.controller(),
+
   actions: {
     saveAccount(){
-      this.transitionToRoute("browse")
+      // After everthying has been loaded, redirect user to requested url
+      var attemptedTransition = this.get('authenticate').get('attemptedTransition');
+      if (attemptedTransition) {
+        this.set('attemptedTransition', null);
+        attemptedTransition.retry();
+      }else{
+        this.transitionToRoute("browse");
+      }
     }
   }
 });
--- a/app/controllers/account_details.js
+++ b/app/controllers/account_details.js
@@ -1,18 +1,54 @@
 import Ember from "ember";
+import AjaxPromise from 'browse/utils/ajax-promise';
+const { getOwner } = Ember;
 
 export default Ember.Controller.extend({
   authenticate:Ember.inject.controller(),
+  messageBox: Ember.inject.service(),
+  organisationId: Ember.computed.alias('model.organisation.id'),
+  organisationsUserId: Ember.computed.alias('model.organisationsUser.id'),
+
+  clearFormData() {
+    this.set('organisationName', "");
+    this.set('firstName', "");
+    this.set('lastName', "");
+    this.set('mobile', "");
+    this.set('email', "");
+    this.set('title', "");
+    this.set('position', "");
+  },
 
   actions: {
-    saveAccount(){
-      // After everthying has been loaded, redirect user to requested url
-      var attemptedTransition = this.get('authenticate').get('attemptedTransition');
-      if (attemptedTransition) {
-        this.set('attemptedTransition', null);
-        attemptedTransition.retry();
-      }else{
-        this.transitionToRoute("browse");
-      }
+    saveAccount() {
+      var loadingView = getOwner(this).lookup('component:loading').append();
+      var mobilePhone = this.get('model.user.mobile');
+      var firstName = this.get('model.user.firstName');
+      var lastName = this.get('model.user.lastName');
+      var organisationId = this.get('organisationId');
+      var position = this.get('model.organisationsUser.position');
+      var email = this.get('model.user.email');
+      new AjaxPromise("/organisations_users", "POST", this.get('session.authToken'), { organisations_user: {
+        organisation_id: organisationId, position: position, user_attributes: { first_name: firstName,
+        last_name: lastName, mobile: mobilePhone, email: email }}}).then(data =>{
+          this.get("store").pushPayload(data);
+          this.clearFormData();
+          var attemptedTransition = this.get('authenticate').get('attemptedTransition');
+          if (attemptedTransition) {
+            this.set('attemptedTransition', null);
+            attemptedTransition.retry();
+          }else{
+            this.transitionToRoute("browse");
+          }
+      }).catch(xhr => {
+        if (xhr.status === 422) {
+          this.get("messageBox").alert(xhr.responseJSON.errors);
+        } else {
+          throw xhr;
+        }
+      })
+      .finally(() =>
+        loadingView.destroy()
+      );
     }
   }
 });
--- a/app/controllers/account_details.js
+++ b/app/controllers/account_details.js
@@ -57,6 +57,10 @@ export default Ember.Controller.extend({
       .finally(() =>
         loadingView.destroy()
       );
+    },
+
+    goToSearchOrg(){
+      this.transitionToRoute("search_organisation");
     }
   }
 });
--- a/app/controllers/account_details.js
+++ b/app/controllers/account_details.js
@@ -3,6 +3,8 @@ import AjaxPromise from 'browse/utils/ajax-promise';
 const { getOwner } = Ember;
 
 export default Ember.Controller.extend({
+  queryParams: ['orgId'],
+
   authenticate:Ember.inject.controller(),
   messageBox: Ember.inject.service(),
   organisationId: Ember.computed.alias('model.organisation.id'),
@@ -18,6 +20,16 @@ export default Ember.Controller.extend({
     this.set('position', "");
   },
 
+  redirectToTransitionOrBrowse() {
+    var attemptedTransition = this.get('authenticate').get('attemptedTransition');
+    if (attemptedTransition) {
+      this.set('attemptedTransition', null);
+      attemptedTransition.retry();
+    }else{
+      this.transitionToRoute("browse");
+    }
+  },
+
   actions: {
     saveAccount() {
       var loadingView = getOwner(this).lookup('component:loading').append();
@@ -27,21 +39,17 @@ export default Ember.Controller.extend({
       var organisationId = this.get('organisationId');
       var position = this.get('model.organisationsUser.position');
       var email = this.get('model.user.email');
+      var title = this.get('model.user.title');
       new AjaxPromise("/organisations_users", "POST", this.get('session.authToken'), { organisations_user: {
         organisation_id: organisationId, position: position, user_attributes: { first_name: firstName,
-        last_name: lastName, mobile: mobilePhone, email: email }}}).then(data =>{
+        last_name: lastName, mobile: mobilePhone, email: email, title: title }}}).then(data =>{
           this.get("store").pushPayload(data);
           this.clearFormData();
-          var attemptedTransition = this.get('authenticate').get('attemptedTransition');
-          if (attemptedTransition) {
-            this.set('attemptedTransition', null);
-            attemptedTransition.retry();
-          }else{
-            this.transitionToRoute("browse");
-          }
+          this.redirectToTransitionOrBrowse();
       }).catch(xhr => {
         if (xhr.status === 422) {
-          this.get("messageBox").alert(xhr.responseJSON.errors);
+          // this.get("messageBox").alert(xhr.responseJSON.errors);
+          this.redirectToTransitionOrBrowse();
         } else {
           throw xhr;
         }
--- a/app/controllers/account_details.js
+++ b/app/controllers/account_details.js
@@ -1,0 +1,9 @@
+import Ember from "ember";
+
+export default Ember.Controller.extend({
+  actions: {
+    saveAccount(){
+      this.transitionToRoute("browse")
+    }
+  }
+});
--- a/app/controllers/search_module.js
+++ b/app/controllers/search_module.js
@@ -1,0 +1,92 @@
+import Ember from "ember";
+import InfinityRoute from "ember-infinity/mixins/route";
+
+export default Ember.Controller.extend(InfinityRoute, {
+
+  queryParams: ["showQuantityItems"],
+  showQuantityItems: false,
+
+  getCurrentUser: Ember.computed(function(){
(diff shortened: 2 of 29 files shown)


app/controllers/search_module.js
```diff
@@ -0,0 +1,92 @@
...
+  }),
+
+  applyFilter() {
+    var searchText = this.get("searchText");
+    if (searchText.length > 0) {
+      this.set("isLoading", true);
+      this.set("hasNoResults", false);
+      if(this.get("unloadAll")) { this.get("store").unloadAll(); }
+
+      this.infinityModel(this.get("searchModelName"),
```
patrixr: @abulsayyad123 Did you get a change to test this infinite scrolling ? 
It seemed to have problems when I tested it on Stock. 
A good way to test is to make smaller pages (e.g 5) and make sure we can scroll to see everything. It would be interesting to see how it behaves here : )

app/controllers/search_module.js
```diff
@@ -0,0 +1,92 @@
...
+        { perPage: 25, startingPage: 1, modelPath: 'filteredResults',stockRequest: true },
+        { searchText: "searchText", itemId: "itemSetId", toDesignateItem: "toDesignateItem", showQuantityItems: "showQuantityItems" })
+        .then(data => {
+          data.forEach(record => {
+            if(record.constructor.toString() === "stock@model:designation:") {
```
patrixr: Don't think this is needed, we discussed it with @Darshan-Chauhan for stock
orders_packages are now available with designations, and we likely want to avoid very specific pieces of code like this in a generic re-usable model

app/controllers/account_details.js
```diff
@@ -0,0 +1,80 @@
...
+      attemptedTransition.retry();
+    }else{
+      this.transitionToRoute("browse");
+    }
+  },
+
+  actions: {
+    saveAccount() {
+      var loadingView = getOwner(this).lookup('component:loading').append();
+      var mobilePhone = this.get('model.user.mobile');
```
Darshan-Chauhan: I think caching `this.get("model.user")` would be a good idea

app/controllers/account_details.js
```diff
@@ -0,0 +1,80 @@
...
+      var firstName = this.get('model.user.firstName');
+      var lastName = this.get('model.user.lastName');
+      var organisationId = this.get('organisationId');
+      var position = this.get('model.organisationsUser.position');
+      var email = this.get('model.user.email');
+      var title = this.get('selectedTitle.name');
+      new AjaxPromise("/organisations_users", "POST", this.get('session.authToken'), { organisations_user: {
+        organisation_id: organisationId, position: position, user_attributes: { first_name: firstName,
```
Darshan-Chauhan: Looks too messy. May be create a function `organisationsUserParams` which returns hash of options?

app/controllers/account_details.js
```diff
@@ -0,0 +1,80 @@
...
+        last_name: lastName, mobile: mobilePhone, email: email, title: title }}}).then(data =>{
+          this.get("store").pushPayload(data);
+          this.clearFormData();
+          this.redirectToTransitionOrBrowse();
+      }).catch(xhr => {
+        if (xhr.status === 422) {
+          // this.get("messageBox").alert(xhr.responseJSON.errors);
+          this.redirectToTransitionOrBrowse();
```
Darshan-Chauhan: Why are we redirecting on 422?

app/controllers/account_details.js
```diff
@@ -0,0 +1,80 @@
...
+        } else {
+          throw xhr;
```
Darshan-Chauhan: Any reason for throwing an exception rather than showing using `messageBox`??

app/models/user.js
```diff
@@ -21,5 +24,9 @@ export default Addressable.extend({
 
   fullName: Ember.computed('firstName', 'lastName', function(){
     return (this.get('firstName') + " " + this.get('lastName'));
+  }),
+
+  isInfoComplete: Ember.computed('firstName', 'lastName', 'email', 'mobile', function(){
+    return this.get('firstName') && this.get('firstName').length !== 0 && this.get('lastName') && this.get('lastName').length !== 0 && this.get('email') && this.get('email').length !== 0 && this.get('mobile') && this.get('mobile').length !== 0 && this.get('title') && this.get('title').length !== 0;
```
patrixr: An empty string evaluates to false, so there is no need to check for the length as well.
You can check that by running `Boolean("") => false`

You could replace 
`this.get('firstName') && this.get('firstName').length !== 0 && ...`
by
`this.get('firstName') && ...`

Although checking the length would be a good practice in most languages, so I'm not explicitly opposed to it, it does show caution which is not a bad thing.

app/routes/post_login.js
```diff
@@ -38,15 +38,36 @@ export default Ember.Route.extend(preloadDataMixin, {
         new AjaxPromise(`/orders/${draftOrder.id}`, "PUT", this.get('session.authToken'), { order: orderParams })
           .then(data => {
             this.get("store").pushPayload(data);
+            this.redirectToTransitionOrDetails();
           });
```
patrixr: I've been seeing a lot of uncaught promises and async calls in Ember. Should we not add a .catch() hook to the promises, how are the errors handled ? 
Unless there is an "ember way" of handling things, I think promises should be caught.

app/controllers/search_module.js
Darshan-Chauhan: @patrixr , also we have `goodcity-lib` addon which we use in Stock and Browse and is specifically designed for reusable code. So I think placing a `search_module` controller in `goodcity-lib` would be a better idea.

What do you think @abulsayyad123  ??

app/controllers/search_module.js
Darshan-Chauhan: @abulsayyad123 , I see infiniteScroll here but I don't see `"ember-infinity"` addon which is needed for infinite scroll. May be we should move `"ember-infinity"` also to `goodcity-lib` with `search_module` controller?

CC: @patrixr

tests/acceptance/account_details-test.js
```diff
@@ -0,0 +1,86 @@
...
+    user = make("user");
+    organisation = make("organisation");
+    organisationsUser = make("organisations_user", {user_id: user.id, organisation_id: organisation.id});
+    pkg = make('package');
+    ordersPackage = make("orders_package", { quantity: 1, state: "requested", package: pkg, packageId: pkg.id, order: order});
+    order = make("order", { state: "draft", created_by_id: user.id });
+    gogo_van = make("gogovan_transport");
+    gcOrganisations = make("gc_organisation", {nameEn: "GCC club"});
+
+    $.mockjax({url: "/api/v1/available_*", type: 'GET', status: 200, responseText:["2018-06-14", "2018-06-15", "2018-06-16", "2018-06-19", "2018-06-20", "2018-06-21"]});
```
patrixr: Mockjax mocks should be cleared after the tests are run, otherwise they add up and might end up conflicting with other tests later. Plus this adds the same mock to the list every time the test is run, which isn't needed.

You can use `$.mockjax.clear()` to clear everything (might affect other tests)
or you can clear mocks individually
`
const mockId = $.mockjax("/api/endpoint", ...);
$.mockjax.clear(mockId)
`

app/controllers/search_organisation.js
```diff
@@ -0,0 +1,46 @@
+import Ember from "ember";
+import searchModule from "./search_module";
+
+export default searchModule.extend({
+  minSearchTextLength: 3,
+
+  onSearchTextChange: Ember.observer("searchText", function(){
```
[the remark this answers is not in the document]
Darshan-Chauhan: Agree with @patrixr , overriding isn't of any use if functionality is not changed/added.

tests/factories/user.js
```diff
@@ -16,6 +16,8 @@ FactoryGuy.define('user',{
     id:        FactoryGuy.generate('id'),
     firstName: FactoryGuy.generate('collectionFirstName'),
     lastName:  FactoryGuy.generate('collectionLastName'),
+    email: "rock.programmer@hotmail.com",
```
patrixr: 🎸 👨‍🎤 🎶

app/locales/zh/translations.js
```diff
@@ -67,22 +68,48 @@ export default {
...
     "login" : "登入",
-    "smscode": "獲取短訊驗證碼"
+    "smscode": "獲取短訊驗證碼",
+    "privacy": "We don't share your data. We only use it to complete your goods request(s). See our ",
+    "policy": "Privacy Policy.",
+    "mobile_warning": "Please provide a valid Hong Kong mobile number"
+  },
+
+  "account": {
+    "account": "Account Details",
```
Darshan-Chauhan: account.account seems ambiguous

patrixr:
Nice job @abulsayyad123 ! 😃 
Added a few comments, just "best practices", nothing fundamentally wrong : )

app/models/organisations_user.js
```diff
@@ -8,5 +9,9 @@ export default Model.extend({
   organisationId:  attr('number'),
   position:        attr('string'),
   user:            belongsTo('user', { async: false }),
-  organisation:    belongsTo('organisation', { async: false })
+  organisation:    belongsTo('organisation', { async: false }),
+
+  isInfoComplete: Ember.computed('position', function(){
+    return this.get('position') && this.get('position').length !== 0;
```
Darshan-Chauhan: Why such computed property is added on model and not in controller??

app/models/user.js
Darshan-Chauhan: No need to compare length with 0. Also why in model and not in controller?

app/routes/account_details.js
```diff
@@ -0,0 +1,22 @@
+import Ember from "ember";
+import AuthorizeRoute from './authorize';
+
+export default AuthorizeRoute.extend({
+  model(params) {
+    return Ember.RSVP.hash({
+      organisation: params.orgId? this.store.peekRecord('gc_organisation', parseInt(params.orgId)) : this.store.peekAll('organisation', {reload: true}).objectAt(0),
```
Darshan-Chauhan: Why `{ reload: true }` is needed here?

app/routes/account_details.js
```diff
@@ -0,0 +1,22 @@
...
+      order: this.store.peekAll("order").filterBy("state", "draft")
+    });
+  },
+
+  setupController(controller, model) {
+    this._super(controller, model);
+    this.controllerFor('application').set('showSidebar', false);
+  },
+
+  deactivate() {
```
Darshan-Chauhan: Good to see use of `deactivate` here 👏

app/routes/post_login.js
```diff
@@ -38,15 +38,36 @@ export default Ember.Route.extend(preloadDataMixin, {
...
       }
+    }else{
```
Darshan-Chauhan: spaces before and after else

app/routes/post_login.js
```diff
@@ -38,15 +38,36 @@ export default Ember.Route.extend(preloadDataMixin, {
...
+    }
+  },
 
-    // After everthying has been loaded, redirect user to requested url
-    var attemptedTransition = this.controllerFor('login').get('attemptedTransition');
-    if (attemptedTransition) {
-      attemptedTransition.retry();
-      this.set('attemptedTransition', null);
+  isDetailsComplete(){
+    var organisation = this.store.peekAll('organisation').objectAt(0);
```
Darshan-Chauhan: How do you know that it'll return the `organisation/organisationsUser/user` that you need? Because you're getting all records from ember-data and taking first object. 
The OrganisationUser you get might not be related to that Organisation at all.


CC: @patrixr , @namrataukirde

app/templates/account_details.hbs
```diff
@@ -0,0 +1,115 @@
...
+        <section class="left-small">
+          {{fa-icon "angle-left"}} {{#link-to "authenticate" classNames="back"}}{{t "back"}}{{/link-to}}
+        </section>
+
+        <section class="middle tab-bar-section">
+          <h1 class="title">{{t "account.account"}}</h1>
+        </section>
+      </nav>
+
+      {{#if model.order.length}}
```
Darshan-Chauhan: Would it be better to check cart rather than draft order?? User's draft Order could be empty also and it'll stil show this

app/templates/account_details.hbs
```diff
@@ -0,0 +1,115 @@
...
+              </div>
+            </div>
+
+            <br>
+
+            {{#validatable-form onSubmit=(action 'saveAccount') as |form|}}
+              <div class="row">
+                {{#form-control}}
+                  <div class="small-12 columns" {{action "goToSearchOrg"}}>
+                      <label class="form-label">{{t "account.organisation"}}</label>
```
Darshan-Chauhan: Indentation

app/templates/account_details.hbs
```diff
@@ -0,0 +1,115 @@
...
+                        {{#if model.organisation}}
+                            {{input type="text" id = "organisation_details" classNames="number_input center_align_placeholder" required='true' value=model.organisation.nameEn}}
```
Darshan-Chauhan: ^

app/templates/account_details.hbs
```diff
@@ -0,0 +1,115 @@
...
+                        {{else}}
+                            {{input type="text" id = "organisation_details" classNames="number_input center_align_placeholder" required='true' }}
```
Darshan-Chauhan: ^

app/templates/account_details.hbs
```diff
@@ -0,0 +1,115 @@
...
+                        {{/if}}
+                  </div>
```
Darshan-Chauhan: ^

app/templates/account_details.hbs
```diff
@@ -0,0 +1,115 @@
...
+                  <div class="small-12 columns input-error warning_empty">{{fa-icon "warning"}}&nbsp;&nbsp;{{t "account.organisation_warning"}}</div>
+                {{/form-control}}
+              </div>
+              <br>
+              <div class="row">
+                {{#form-control}}
+                  <div class="small-2 columns user_title">
+                      <label class="form-label">{{t "account.title"}}</label>
```
Darshan-Chauhan: ^

tests/acceptance/account_details-test.js
```diff
@@ -0,0 +1,86 @@
...
+import { module, test } from 'qunit';
+import startApp from '../helpers/start-app';
+import {make} from 'ember-data-factory-guy';
+import { mockFindAll } from 'ember-data-factory-guy';
+
+var App, user, organisation, organisationsUser, gogo_van, order, pkg, ordersPackage, gcOrganisations;
+
+module('Acceptance | Account Details Page', {
+  beforeEach: function() {
+    window.localStorage.authToken = '"eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1c2VyX2lkIjo2LCJpYXQiOjE1MTg3NzI4MjcsImlzcyI6Ikdvb2RDaXR5SEsiLCJleHAiOjE1MTk5ODI0Mjd9.WdsVvss9khm81WNScV5r6DiIwo8CQfHM1c4ON2IACes"';
```
Darshan-Chauhan: Why do we need authToken here  when `startApp` function used below already lets you log in?

tests/acceptance/account_details-test.js
```diff
@@ -0,0 +1,86 @@
...
+        organisations: [organisation.toJSON({includeId: true})],
+        organisations_users: [organisationsUser.toJSON({includeId: true})]
+      }});
+
+    $.mockjax({url: "api/v1/gc_org*", type: 'GET', responseText: {
+       meta :{ total_pages:1, search: "club"},
+       gc_organisations: [gcOrganisations.toJSON({includeId: true}), organisation.toJSON({includeId: true})],
+       organisations: [gcOrganisations.toJSON({includeId: true}), organisation.toJSON({includeId: true})],
+    }});
+    mockFindAll('order').returns({ json: {orders: [order.toJSON({includeId: true})], packages: [pkg.toJSON({includeId: true})], orders_packages: [ordersPackage.toJSON({includeId: true})]}});
```
Darshan-Chauhan: May be writing response for different model in different lines will make it more readable?

Darshan-Chauhan:
Good work @abulsayyad123 , added some comments. :)

app/controllers/search_module.js
patrixr: That's a good idea

app/models/user.js
abulsayyad123: I thought  model state information should be within model itself. Had tried with this.get('firstName') only but it didn't work. So , had to explicitly add length check.

app/templates/account_details.hbs
abulsayyad123: When draft order has no packages it gets deleted.

abulsayyad123:
I found few h-cases regarding it and I am working on it. Thus closing this PR for now. 

Thanks.

app/controllers/search_module.js
abulsayyad123: That is what I have plan of putting this search_module and search_organisation controller in goodcity_lib.

PR #208, opened 2018-09-27
Fix banner images on cordova

patrixr:
Path for images should be different on cordova.

Also noticed the session.language was not properly initialised, which sometimes caused weird behaviors with the language switcher component. Added a couple of lines to initialise it.


app/templates/components/logo-banner.hbs
```diff
@@ -5,8 +5,12 @@
             <div class="contributor jockey_club"> {{ t "banner.funded_by" }} </div>
         </div>
         <div class="contributors logos">
-            <div class="contributor logo crossroads"></div>
-            <div class="contributor logo jockey_club"></div>
+            <div class="contributor crossroads">
```
Darshan-Chauhan: indentation 😄

patrixr: Every time I configure my editor, it resets itself  😞

PR #209, opened 2018-10-03
Gcw 2132 -New user sign up feature along with account details page added 

abulsayyad123:
Hi Team,
This is PR for New User Sign up feature. https://jira.crossroads.org.hk/browse/GCW-2132.
And some other changes added which includes.

User account details page added.
Search GC organisation page added for account details page.
User redirection to browse and account-details page according to completion of user details.
Test cases added.
New styling added according to new mockup design.
Please review this PR and let me know if any change is required.

Thanks


app/models/gc_organisation.js
```diff
@@ -0,0 +1,16 @@
+import Model from 'ember-data/model';
```
patrixr: @Darshan-Chauhan @abulsayyad123 @namrataukirde Just a general question. Would it make sense to have the models in goodcity-lib as well ? Since they are shared across apps.

app/routes/search_organisation.js
```diff
@@ -0,0 +1,3 @@
+import AuthorizeRoute from './authorize';
```
Darshan-Chauhan: Move this to goodcity-lib

app/templates/account_details.hbs
```diff
@@ -0,0 +1,120 @@
...
+                  </div>
+                {{/form-control}}
+              </div>
+              <br>
+              <div class="row">
+                {{#form-control}}
+                  <div class="small-12 columns">
+                    <label class="form-label">{{t "account.position"}}</label>
+                    {{#if model.organisationsUser}}
+                        {{input type="text" value=model.organisationsUser.position id="position" classNames="number_input center_align_placeholder" required='true' autoFocus=true}}
```
Darshan-Chauhan: indentation

app/templates/login.hbs
```diff
@@ -12,22 +22,31 @@
                 <div class="small-12 columns">
                  <label class="form-label">{{t "login.hk_only"}}</label>
                 </div>
+                <div class="small-12 columns">
+                 <label class="form-label">{{t "login.sms_pincode"}}</label>
```
Darshan-Chauhan: indentation

app/templates/login.hbs
```diff
@@ -12,22 +22,31 @@
...
-                <div class="small-10 medium-8 columns">
-                  {{#form-control}}
-                    <span class="prefix">+852</span>
-                    {{numeric-input value=mobilePhone name="mobilePhone" id="mobile" maxlength="8" required=true pattern=".{8,}" autoFocus=true}}
-                  {{/form-control}}
-                </div>
+              <div class="row">
+                {{#form-control}}
+                  <div class="small-12 medium-8 columns">
+                      <span class="prefix">+852</span>
```
Darshan-Chauhan: indentation

app/templates/search_organisation.hbs
```diff
@@ -0,0 +1,65 @@
...
+            name="searchText"
+            id="searchText"
+            value=searchText
+            disabled=false}}
+
+          {{#if hasSearchText}}
+            <i class="fa fa-times-circle" {{action "clearSearch"}}></i>
+          {{/if}}
+        </div>
+        <div class="large-1 columns"></div>
```
Darshan-Chauhan: indentation

app/models/gc_organisation.js
patrixr: Agreed, I don't think it's urgent. We can look at refactoring when there's a bit of downtime

Darshan-Chauhan: Yes but I think we can move this model and route to `goodcity-lib` as it is part of this PR. 
For others we can create a ticket.

@abulsayyad123 , can you move this model to `goodcity-lib`?

Darshan-Chauhan:
@abulsayyad123 , good work. Added some comments

app/routes/account_details.js
```diff
@@ -0,0 +1,27 @@
...
+  model(params) {
+    var user = this.get('session.currentUser');
+    var organisationsUser = user.get('organisationsUsers') && user.get('organisationsUsers.firstObject');
+    var organisation = organisationsUser && organisationsUser.get('organisation');
+
+    return Ember.RSVP.hash({
+      organisation: params.orgId? this.store.peekRecord('gc_organisation', parseInt(params.orgId)) : organisation,
+      organisationsUser: organisationsUser,
+      user: user,
+      order: this.store.peekAll("order").filterBy("state", "draft")
```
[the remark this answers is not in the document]
abulsayyad123: We have the feature that if there is no packages in order in `draft` state the `Order` gets deleted.

app/models/gc_organisation.js
abulsayyad123: Sure, will move model and route to goodcity-lib.

app/routes/account_details.js
```diff
@@ -0,0 +1,27 @@
+import Ember from "ember";
+import AuthorizeRoute from './authorize';
+
+export default AuthorizeRoute.extend({
+  model(params) {
+    var user = this.get('session.currentUser');
+    var organisationsUser = user.get('organisationsUsers') && user.get('organisationsUsers.firstObject');
```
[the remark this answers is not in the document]
abulsayyad123: No they are same user, I have just taken them in separated variables for simplicity and readability.

app/models/gc_organisation.js
abulsayyad123: There, is problem to add `gc_organisation` model in goodcity-lib  right now because there is association for `designation` which exists in Stock but not in Browse app. And browse uses `order` instead of  `designation`. We can create seperate ticket for refactoring code.

patrixr:
Well done !

app/controllers/account_details.js
```diff
@@ -0,0 +1,111 @@
...
+  organisationId: Ember.computed.alias('model.organisation.id'),
+  organisationsUserId: Ember.computed.alias('model.organisationsUser.id'),
+  position: "",
+
+  userTitle: Ember.computed('model', function() {
+    let userTitle = this.get('model.user.title');
+    let titles = this.get('titles');
+
+    if(userTitle) {
+      let filteredUserTitle = titles.filter((item) => userTitle === item.id);
```
Darshan-Chauhan: titles.filter((item) => userTitle === item.id); can we change `item`to `title`?

app/controllers/account_details.js
```diff
@@ -0,0 +1,111 @@
...
+      return { name: filteredUserTitle[0].name.string, id: userTitle };
+    } else {
+      return { name: titles.get('firstObject.name').string, id: 'Mr' };
```
Darshan-Chauhan: I think this is the default condition, is no title exist then we show `Mr`, right?
If that so than I think we can keep it outside else block as default. 
AS else block statement won't get executed because if condition has return statement

app/controllers/account_details.js
```diff
@@ -0,0 +1,111 @@
...
+    }
+  }),
+
+  selectedTitle: Ember.computed('userTitle', function (){
+    return{ name: this.get('userTitle.name'), id: this.get('userTitle.id')};
```
Darshan-Chauhan: space between return and `{`

app/controllers/account_details.js
```diff
@@ -0,0 +1,111 @@
...
+  }),
+
+  titles: Ember.computed(function() {
+    let mr = this.get("i18n").t("account.user_title.mr");
```
Darshan-Chauhan: cache `this.get("i18n")`

app/controllers/account_details.js
```diff
@@ -0,0 +1,111 @@
...
+      }
+    };
+    if (organisationsUserId) {
+      params.id = organisationsUserId;
+    }
+
+    return params;
+  },
+
+  saveOrUpdateAccount(url, actionType) {
```
Darshan-Chauhan: This should be an action instead of function.
Function should be only written to help actions execute.
Functions shouldn't fire requests to API, that's what actions are for.

app/models/user.js
```diff
@@ -21,5 +24,9 @@ export default Addressable.extend({
 
   fullName: Ember.computed('firstName', 'lastName', function(){
     return (this.get('firstName') + " " + this.get('lastName'));
+  }),
+
+  isInfoComplete: Ember.computed('firstName', 'lastName', 'email', 'mobile', function(){
+    return this.get('firstName') && this.get('firstName').length && this.get('lastName') && this.get('lastName').length && this.get('email') && this.get('email').length && this.get('mobile') && this.get('mobile').length && this.get('title') && this.get('title').length;
```
Darshan-Chauhan: Can we refactor this??
May be write a function which checks values for attributes exists or not and returns true or false??
May be something like this would work

```
hasValue(...args) {
  args.forEach(arg => {
    if(!this.get("arg") || !this.get("arg").length) {
      return false;
    }
  })
}
```

app/templates/account_details.hbs
```diff
@@ -0,0 +1,120 @@
...
+      {{#if model.order.length}}
+        {{small-cart-toggle-view}}
+      {{/if}}
+
+      <div class="row">
+        <div class="small-12 columns">
+          <div class="mobile center_align">
+            <div class="row">
+              <div class="small-12 columns">
+               <label class="form-label">{{t "account.welcome"}}</label>
```
Darshan-Chauhan: is this 1 space indentation??

app/templates/account_details.hbs
```diff
@@ -0,0 +1,120 @@
...
+              </div>
+              <div class="small-12 columns">
+               <label class="form-label">{{t "account.contact_details"}}</label>
```
Darshan-Chauhan: ^

app/models/organisations_user.js
```diff
@@ -8,5 +9,9 @@ export default Model.extend({
   organisationId:  attr('number'),
   position:        attr('string'),
   user:            belongsTo('user', { async: false }),
-  organisation:    belongsTo('organisation', { async: false })
+  organisation:    belongsTo('organisation', { async: false }),
+
+  isInfoComplete: Ember.computed('position', function(){
+    return this.get('position') && this.get('position').length !== 0 ;
```
namrataukirde: What if we use `!this.get('position').length` instead of `this.get('position').length !== 0` ? Will that work ?

app/models/user.js
```diff
@@ -21,5 +24,20 @@ export default Addressable.extend({
...
+  }),
+
+  isInfoComplete: Ember.computed('firstName', 'lastName', 'email', 'mobile', 'title', function(){
+    return this.hasValue('firstName', 'lastName', 'email', 'mobile', 'title');
+  }),
+
+  hasValue(...args) {
+    let hasValue = true;
+    for (let i = 0; i < args.length; i ++) {
+      if(!this.get(args[i]) || !this.get(args[i]).length) {
```
namrataukirde: What will happen to `this.get(args[i])` if consider first_name value will be blank string ? 🤔

namrataukirde: This code look reusable and not related to specific model. 

What if we move this code to utility_methods file ?

CC: @Darshan-Chauhan

app/routes/account_details.js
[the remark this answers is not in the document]
namrataukirde: I feel we have that feature that's fine but still in code we should try to avoid such case and change code as @Darshan-Chauhan suggested. :)

Let me know your thoughts.

CC: @Darshan-Chauhan

app/styles/templates/_login.scss
```diff
@@ -4,10 +4,46 @@
   }
 }
 
-.login-page {
-  @media #{$medium-only}{
-    margin-left: 3.5rem;
+.warning_empty{
+  padding-left: 1rem !important;
```
namrataukirde: All those `!important` are really needed ?

There are lots of `!important` in this file. Just verify it once and if not needed then remove. So many `important` creates mess later on. You know that already. So just confirming. :)

app/templates/search_organisation.hbs
```diff
@@ -0,0 +1,65 @@
...
+          {{/infinity-loader}}
+
+        {{else if hasNoResults}}
+          <div class='no_result'>{{t "no_search_results"}}</div>
+        {{/if}}
+      {{/if}}
+    </div>
+
+  </div>
+  {{!-- {{partial "sticky_footer"}} --}}
```
namrataukirde: Remove this if not needed.

tests/factories/user.js
```diff
@@ -16,6 +16,8 @@ FactoryGuy.define('user',{
     id:        FactoryGuy.generate('id'),
     firstName: FactoryGuy.generate('collectionFirstName'),
     lastName:  FactoryGuy.generate('collectionLastName'),
+    email: "rock.programmer@hotmail.com",
```
namrataukirde: We can use faker for creating data in future may be like we do in api.

https://github.com/johno/ember-faker

We don't need to do this in PR.

Adding this comment just for our attention. :)

CC: @steveyken  @Darshan-Chauhan  @patrixr

namrataukirde:
Good Work @abulsayyad123 , Just few comments to clarify my queries.. :)

app/models/user.js
namrataukirde: @Darshan-Chauhan , Its use at just one place.. Right?

PR #214, opened 2018-10-17
Fixed:GCW 2153- display inventory number added

abulsayyad123:
Hi Team,
               This is PR for GCW-2153- adding inventory number on browse to redirect user to stock app on inventory number Item for Order Fulfilment role user.
https://jira.crossroads.org.hk/browse/GCW-2153

- I have added the fix and also added test case for the same.
- Will be creating separate ticket for the mobile deep linking i.e app to app navigation.
Let me know if any change is required.

Thanks

### DIFF
--- a/app/controllers/item.js
+++ b/app/controllers/item.js
@@ -1,4 +1,5 @@
 import Ember from 'ember';
+import config from "../config/environment";
 
 export default Ember.Controller.extend({
 
@@ -37,7 +38,7 @@ export default Ember.Controller.extend({
   }),
 
   hasDraftOrder: Ember.computed.alias("session.draftOrder"),
-  isCharityUser: Ember.computed.alias("session.isCharityUser"),
+  isOrderFulfilmentUser: Ember.computed.alias("session.currentUser.isOrderFulfilmentUser"),
 
   presentInCart: Ember.computed('item', 'cart.counter', function(){
     return this.get('cart').hasCartItem(this.get('item'));
@@ -84,19 +85,8 @@ export default Ember.Controller.extend({
       this.set('previewUrl', image.get("previewImageUrl"));
     },
 
-    gotoStockItem(inventoryNumber) {
-      let url = window.location.hostname.split('.')[0];
-      let finalUrl;
-      switch(url) {
-          case "browse-staging":
-              finalUrl = "https://stock-staging.goodcity.hk/items/"+inventoryNumber;
-              break;
-          case "browse":
-              finalUrl = "https://stock.goodcity.hk/items/"+inventoryNumber;
-              break;
-          case "localhost":
-              finalUrl = "http://localhost:4203/items/"+inventoryNumber;
-      }
+    goToStockItem(inventoryNumber) {
+      let finalUrl = config.APP.BASE_HOST_URL + "/items/" + inventoryNumber;
       window.open(finalUrl, '_blank');
     },
 
--- a/app/controllers/item.js
+++ b/app/controllers/item.js
@@ -37,6 +37,7 @@ export default Ember.Controller.extend({
   }),
 
   hasDraftOrder: Ember.computed.alias("session.draftOrder"),
+  isCharityUser: Ember.computed.alias("session.isCharityUser"),
 
   presentInCart: Ember.computed('item', 'cart.counter', function(){
     return this.get('cart').hasCartItem(this.get('item'));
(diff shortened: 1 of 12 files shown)


app/controllers/item.js
```diff
@@ -81,6 +83,23 @@ export default Ember.Controller.extend({
       this.set('previewUrl', image.get("previewImageUrl"));
     },
 
+    gotoStockItem(inventoryNumber) {
+      let environment = config.staging ? "staging" : config.environment;
+      let finalUrl;
+
+      switch(environment) {
```
Darshan-Chauhan: Refactor switch case to use **config.APP.BASE_HOST_URL.**
I don't think we need to compare environments here.

app/controllers/item.js
```diff
@@ -81,6 +83,23 @@ export default Ember.Controller.extend({
...
+        case "staging":
+          finalUrl = config.APP.BASE_HOST_URL+"/items/"+inventoryNumber;
```
Darshan-Chauhan: Also add space before and after "+"

app/locales/en/translations.js
```diff
@@ -123,7 +123,8 @@ export default {
     "quantity": "Quantity",
     "size": "Size",
     "categories": "Categories",
-    "image_unavailable": "No image available."
+    "image_unavailable": "No image available.",
+    "inventory": "Inventory Number"
```
Darshan-Chauhan: Will `inventory_number` look much readable here?

app/controllers/item.js
```diff
@@ -81,6 +83,23 @@ export default Ember.Controller.extend({
       this.set('previewUrl', image.get("previewImageUrl"));
     },
 
+    gotoStockItem(inventoryNumber) {
```
Darshan-Chauhan: A small change **goToStockItem**

app/services/session.js
```diff
@@ -13,6 +13,10 @@ export default Ember.Service.extend({
     return store.peekAll('user').get('firstObject') || null;
   }).volatile(),
 
+  isOrderFulfilmentUser: Ember.computed(function() {
```
Darshan-Chauhan: This should be method on User model which gives filtered User based on role

app/services/session.js
```diff
@@ -13,6 +13,10 @@ export default Ember.Service.extend({
...
+    return this.get('store').peekAll('role').getEach('name').includes('Order fulfilment');
```
Darshan-Chauhan: `includes` doesn't work with IE. Use `indexOf`

app/models/user.js
```diff
@@ -21,5 +21,9 @@ export default Addressable.extend({
 
   fullName: Ember.computed('firstName', 'lastName', function(){
     return (this.get('firstName') + " " + this.get('lastName'));
+  }),
+
+  isOrderFulfilmentUser: Ember.computed('roles', function() {
```
Darshan-Chauhan: If we want to check whether User has some particular role or not we'll have to create computed property for each role... And I think we can refactor it.

May be creating a static function in User model called "hasRole" which takes user object as first argument and role name(string) as second argument and returns truthy value depending or role exists or not.

This should work like rails scopes.

CC: @namrataukirde

app/controllers/item.js
```diff
@@ -81,6 +83,11 @@ export default Ember.Controller.extend({
       this.set('previewUrl', image.get("previewImageUrl"));
     },
 
+    goToStockItem(inventoryNumber) {
+      let finalUrl = config.APP.BASE_HOST_URL + "/items/" + inventoryNumber;
```
Darshan-Chauhan: Do we really need this variable here??

abulsayyad123: Passing long string as a parameter looks ugly, so assigned it to a variable.

config/environment.js
```diff
@@ -81,14 +81,15 @@ module.exports = function(environment) {
     // RESTAdapter Settings
     ENV.APP.SOCKETIO_WEBSERVICE_URL = 'http://localhost:1337/goodcity';
     ENV.APP.API_HOST_URL = 'http://localhost:3000';
+    ENV.APP.BASE_HOST_URL = 'http://localhost:4203';
```
Darshan-Chauhan: Can we change this name to `STOCK_APP_HOST_URL` for readability and clarity purposes?

abulsayyad123: Yes, that is more clear.

app/models/user.js
```diff
@@ -39,5 +39,17 @@ export default Addressable.extend({
       }
     }
     return hasValue;
+  },
+
+  //*****************
```
patrixr: I like those comments :)

PR #223, opened 2018-11-05
Fixed: Inventory Number display issue on Logged out state.

abulsayyad123:
Hi Team,
This is Fix for the Inventory number display issue in logged out state.
https://jira.crossroads.org.hk/browse/GCW-2153?focusedCommentId=31806&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-31806.

Please review the PR and let me know if any change is required.
Thanks.

### DIFF
--- a/app/controllers/item.js
+++ b/app/controllers/item.js
@@ -17,7 +17,6 @@ export default Ember.Controller.extend({
   itemNotAvailableShown: false,
   hasCartItems: Ember.computed.alias('application.hasCartItems'),
 
-  
   direction: null,
 
   hasQuantityAndIsAvailable: Ember.observer('item.isAvailable', 'item.packages.@each.orderId', 'item.isUnavailableAndDesignated', function() {
@@ -38,9 +37,10 @@ export default Ember.Controller.extend({
   }),
 
   hasDraftOrder: Ember.computed.alias("session.draftOrder"),
-  isOrderFulfilmentUser: Ember.computed(function() {
+  isOrderFulfilmentUser: Ember.computed('session.authToken', function() {
     let user = this.get('session.currentUser');
-    return user.hasRole('Order fulfilment');
+    let isLoggedIn = this.get('session.isLoggedIn');
+    return user && isLoggedIn ? user.hasRole('Order fulfilment') : false;
   }),
 
   presentInCart: Ember.computed('item', 'cart.counter', function(){
--- a/app/controllers/item.js
+++ b/app/controllers/item.js
@@ -37,10 +37,9 @@ export default Ember.Controller.extend({
   }),
 
   hasDraftOrder: Ember.computed.alias("session.draftOrder"),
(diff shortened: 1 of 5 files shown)


app/controllers/item.js
```diff
@@ -38,9 +37,10 @@ export default Ember.Controller.extend({
   }),
 
   hasDraftOrder: Ember.computed.alias("session.draftOrder"),
-  isOrderFulfilmentUser: Ember.computed(function() {
+  isOrderFulfilmentUser: Ember.computed('session.authToken', function() {
     let user = this.get('session.currentUser');
-    return user.hasRole('Order fulfilment');
+    let isLoggedIn = this.get('session.isLoggedIn');
+    return user && isLoggedIn ? user.hasRole('Order fulfilment') : false;
```
Darshan-Chauhan: I think we can improve the condition a bit.
User can't exist without logging in. So if User object is there that means he/she is logged in or is isLoggedIn is true that means User exists.

Darshan-Chauhan: Also `isOrderFulfilmentUser` computed property should only check is user has `Order fulfilment` role or not. It shouldn't check if User is logged in or not.

app/controllers/item.js
```diff
@@ -38,9 +37,9 @@ export default Ember.Controller.extend({
   }),
 
   hasDraftOrder: Ember.computed.alias("session.draftOrder"),
-  isOrderFulfilmentUser: Ember.computed(function() {
+  isOrderFulfilmentUser: Ember.computed('session.authToken', function() {
```
Darshan-Chauhan: This comouted property should only get executed if user is available. As I've said before it should only check whether User has order fulfillment role or not. It shouldn't check if the user is logged in or not

Darshan-Chauhan:
Also remove package lock file and add yarn.lock file

app/controllers/item.js
```diff
@@ -38,9 +37,9 @@ export default Ember.Controller.extend({
...
     let user = this.get('session.currentUser');
-    return user.hasRole('Order fulfilment');
+    return user && user.hasRole('Order fulfilment');
```
Darshan-Chauhan: Unnecessary conditions

app/controllers/item.js
abulsayyad123: If it is not executed after logged out, then 'Inventory number' remains visible on the screen.

app/controllers/item.js
abulsayyad123: If this is executing after logout than this condition is required. Because it tries to fetch `roles` from `user`.

Darshan-Chauhan: Computed property shouldn't get called if no user exist

Darshan-Chauhan:
@abulsayyad123 clear everything (current user and Everything) in session service's clear method.

PR #236, opened 2018-11-22
Column changes for book appointment

namrataukirde:


app/styles/components/_add-request.scss
```diff
@@ -1,14 +1,14 @@
...
 }
 
 .request-section {
   .request-block {
     border-bottom: 1px solid;
   }
-  .req-description input{
-    margin-top: 1rem !important;
+  .req-description input {
+    margin-top: 0.1rem !important;
```
patrixr: Quick tip to avoid `!important`
CSS gives priority to the rules which are the most precise.
so sometimes adding element type and other classes to the rule allows to avoid using `!important`
For e.g this might work :
```scss
div.some-other-class.req-description input { }
```

Not a change request, just mentioning it 😄

namrataukirde: Yeah.. Here we need important was needed here. But will try nesting it different way. Thanks for input :)

patrixr: No need to nest differently, just adding "div" at the front or adding a class might work, without changing the html :)

namrataukirde: Yeah.. Actually i dont need `important`.  so I removed it. Thanks for flagging it 👍

PR #252, opened 2018-12-04
disable ios build

namrataukirde:
Hi Team,

Disabling ios build for now. We can enable it whenever its actually needed and again we can disable it until we find solution for billing and extra build time 

Thanks.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -338,9 +338,9 @@ workflows:
             filters:
               branches:
                 only: /^(master|live)$/
-        - ios_build_and_deploy:
-            requires:
-              - test_cases
-            filters:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -338,9 +338,9 @@ workflows:
...
-              - test_cases
-            filters:
-              branches:
-                only: /^(master|live)$/
+        # - ios_build_and_deploy:
+        #     requires:
+        #       - test_cases
+        #     filters:
+        #       branches:
+        #         only: /^(master|live)$/
```
Darshan-Chauhan: I think only removing master should've been fine :)

PR #264, opened 2018-12-07
GCW-2209 Hide app version from web app

abhinav098:
Hi team, while testing, found out, app version was displayed on web app also, so made some changes to hide it from web. Please review and comment.
Thanks.

### DIFF
--- a/app/controllers/application.js
+++ b/app/controllers/application.js
@@ -5,7 +5,9 @@ const { getOwner } = Ember;
 
 export default Ember.Controller.extend({
 
-  appVersion: config.APP.VERSION,
+  appVersion: Ember.computed(function(){
+    return config.cordova.enabled ? config.APP.VERSION : null;
+  }),
   subscription: Ember.inject.controller(),
   messageBox: Ember.inject.service(),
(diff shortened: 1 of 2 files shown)


app/controllers/application.js
```diff
@@ -5,7 +5,9 @@ const { getOwner } = Ember;
 
 export default Ember.Controller.extend({
 
-  appVersion: config.APP.VERSION,
+  appVersion: Ember.computed(function(){
```
Darshan-Chauhan: instead this add `isMobileApp: config.cordova.enabled`

app/templates/sidebar.hbs
```diff
@@ -77,7 +77,9 @@
           {{/link-to}}
         </div>
       </div>
-      <div class="line">{{t "application.sidebar.version" }} {{ appVersion }}</div>
+      {{#if appVersion}}
```
Darshan-Chauhan: Here {{#if isMobileApp}}

app/controllers/application.js
abhinav098: changes are already done.

PR #283, opened 2019-01-04
Workaround for issue where addressId was undefined

patrixr:
Hi team, 

One user of Browse had an issue where he/she couldn't proceed pas the transport details page.

After testing I noticed the error comes from tying to save the contact address to the order :

`this.get('order.orderTransport.contact.address.id');` would return undefined _even though_ the data was returned by the server correctly.
I suspect if might be related to those issues :
- https://github.com/emberjs/data/issues/2496
- https://github.com/emberjs/data/issues/5313

why we are seeing it only now, I'm not sure  ¯\\_(ツ)_/¯

I made a quick fix just to get it working for prod, but I am **not** happy with this 😄 😄 

--> any suggestions ? <--

I'm having a hard time remembering why we had to store the address on the order. If it is required, then couldn't `order.addressId` be computed on the server ? to be an alias to `order.orderTransport.contact.address.id` ?

### DIFF
--- a/app/controllers/order/transport_details.js
+++ b/app/controllers/order/transport_details.js
@@ -1,6 +1,7 @@
 import Ember from 'ember';
 import AjaxPromise from './../../utils/ajax-promise';
 import applicationController from './../application';
+import _ from 'lodash';
 const { getOwner } = Ember;
 
 export default applicationController.extend({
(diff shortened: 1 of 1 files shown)


Darshan-Chauhan:
@patrixr , we're storing address with this relationship before `order.orderTransport.contact.address.id` so I created a ticket to store the address direct in `order` model, thus `address_id` in `order model`. 

I suspect we've only added the column but didn't make the changes to use that column. 
I think we'll have to create a ticket to make changes and to correct existing data.

CC: @abhinav098 , @abulsayyad123 , @namrataukirde

Darshan-Chauhan:
@patrixr , https://github.com/crossroads/browse.goodcity/blob/master/app/templates/_confirm_order.hbs#L31-L32

I believe this was the reason we wanted to store address directly to order but I suspect respected change weren't made on front-end side

patrixr:
@Darshan I think I have made the changes. The issue here comes from _reading_ from the `orderTransport.contact.address`  , so it happens before even using the order's address.
Data looks good coming from the server, but Ember is not resolving the relationship

PR #289, opened 2019-01-14
send appropriate booking_type with order request instead of order_tra…

namrataukirde:
Hi Team, 

This PR fixes appointment_type as per association changes on api.

https://jira.crossroads.org.hk/browse/GCW-2315

Please review.

Thanks.

### DIFF
--- a/app/controllers/request_purpose.js
+++ b/app/controllers/request_purpose.js
@@ -53,7 +53,8 @@ export default Ember.Controller.extend(cancelOrder, {
         purpose_ids: purposeIds,
         order_type: 'appointment',
         people_helped: this.get('peopleCount'),
-        district_id: this.get('selectedDistrict.id')
+        district_id: this.get('selectedDistrict.id'),
+        booking_type_id: this.store.peekAll('booking_type').filterBy('nameEn', 'appointment').get('firstObject.id')
       };
 
       let order = this.get('model');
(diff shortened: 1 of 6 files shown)


app/controllers/request_purpose.js
```diff
@@ -53,7 +53,8 @@ export default Ember.Controller.extend(cancelOrder, {
         purpose_ids: purposeIds,
         order_type: 'appointment',
         people_helped: this.get('peopleCount'),
-        district_id: this.get('selectedDistrict.id')
+        district_id: this.get('selectedDistrict.id'),
+        booking_type_id: this.store.peekAll('booking_type').filterBy('nameEn', 'appointment').get('firstObject.id')
```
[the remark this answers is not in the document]
namrataukirde: Yeah.. Good point.. Thanks. We have identifier.. its good to use identifier instead of `nameEn`.. Changing it 👍

app/models/order.js
```diff
@@ -25,6 +25,7 @@ export default Model.extend({
   peopleHelped: attr('number'),
   goodcityRequests:   hasMany('goodcity_request', { async: false }),
   district: belongsTo('district', {async: false}),
+  bookingType:   belongsTo('booking_type', { async: false }),
```
patrixr: if we're adding this to the order model, I assume we need to remove it from the order_transport model. I think that's missing :)

namrataukirde: Yeah @patrixr .. removed it.. forgot to commit that file.. Thanks 😅

patrixr: Cool :D

app/controllers/request_purpose.js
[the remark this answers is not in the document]
namrataukirde: @Darshan-Chauhan did change.. Please have a look.

app/models/order.js
namrataukirde: Committed change 👍

PR #296, opened 2019-01-25
Fix wrong filter for appointments

patrixr:


namrataukirde:
Is this only one or is there anything else which we need to look ?

namrataukirde:
Thanks 👍

PR #301, opened 2019-02-13
GCW-2323 Consolidate online-order and appointment booking flows

patrixr:
Overall changes : 

* Removed the following pages  _(no new page was added)_
  * `transport_details`
  * `order_details`
  * `confirm`
* Renamed `appointment_details` to `schedule_details`. A more generic name as it's used for both types now
* Rewired the navigation between pages (might benefit from using history.back() as well, looks like overcomplicated logic right now)
* The `schedule_details` will show a different input for the timeslot based on the type of order
* Added missing chinese translations

Flow : 

screenshot 2019-02-13 at 12 20 26 pm


patrixr:
_(Looking into code climate issues)_

PR #302, opened 2019-02-13
GCW-2335 basic setup for chat

abhinav098:
Hello team, these are the browse side changes for chat between browse and stock for orders.

### DIFF
--- a/app/controllers/my_orders.js
+++ b/app/controllers/my_orders.js
@@ -36,7 +36,7 @@ export default applicationController.extend({
     var object, value;
 
     content.forEach(function (item) {
-      value = item.get ? item.get(key) : item[key];
+      value = item.get(key);
       object = result.findBy('value', value);
       if (!object) {
         object = {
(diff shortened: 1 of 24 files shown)


app/models/order.js
```diff
@@ -20,7 +21,12 @@ export default Model.extend({
   detailType:       attr('string'),
   districtId:       attr('number'),
   ordersPurposes:     hasMany('ordersPurpose', { async: false }),
+<<<<<<< HEAD
   beneficiaryId: attr('string'),
+=======
```
steveyken: need to resolve this

abhinav098: Yupp, my bad, missed this. will do it right away. Thanks @steveyken

app/controllers/my_orders.js
```diff
@@ -17,6 +18,37 @@ export default applicationController.extend({
...
+    return this.get("noMessage") && !this.get("disabled");
+  }),
+
+  sortedMessages: Ember.computed.sort("selectedOrder.messages", "messageSortOptions"),
+
+  groupedMessages: Ember.computed("sortedMessages", function () {
+    return this.groupBy(this.get("sortedMessages"), "createdDate");
+  }),
+
+  groupBy: function (content, key) {
```
patrixr: note: we have lodash installed as a dependency, feel free to use it (e.g `_.groupBy(...)`)   😄

Darshan-Chauhan: Good suggestion

abhinav098: sure @patrixr , i'll try to implement that way.

PR #305, opened 2019-02-20
Fix redirection from client_information

patrixr:
Somehow there was the following condition : 
```javascript
if (isClientInformation) {
          this.transitionToRoute('order.goods_details', orderId, { queryParams: { fromClientInformation: true }});
}
```

which doesn't make sense, I suspect it's during a merge/conflict that it stayed there. This PR just removes it (and changes the title of a page as per requirements)

### DIFF
--- a/app/controllers/order/client_information.js
+++ b/app/controllers/order/client_information.js
@@ -132,7 +132,7 @@ export default Ember.Controller.extend(cancelOrder, {
         });
     },
 
-    redirectToGoodsDetails(isClientInformation=false) {
+    redirectToGoodsDetails() {
       let order = this.get("order");
       let orderId = order.id;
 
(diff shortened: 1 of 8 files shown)


Darshan-Chauhan:
@patrixr , I think that argument is used here => https://github.com/crossroads/browse.goodcity/blob/7713072ab2a63edf2ed4090a098915d8649d814d/app/controllers/order/client_information.js#L123

patrixr:
> @patrixr , I think that argument is used here =>
> 
> [browse.goodcity/app/controllers/order/client_information.js](https://github.com/crossroads/browse.goodcity/blob/7713072ab2a63edf2ed4090a098915d8649d814d/app/controllers/order/client_information.js#L123)
> 
> Line 123 in [7713072](/crossroads/browse.goodcity/commit/7713072ab2a63edf2ed4090a098915d8649d814d)
> 
>  this.send('redirectToGoodsDetails', true);

But I don't think it's needed, the fact that it is for a beneficiary or not should not influence the navigation. Or am I missing something ? (ping @Darshan-Chauhan )

Darshan-Chauhan:
@patrixr , I agree with you. 
But it's still better to ask @abulsayyad123 or @namrataukirde  who wrote this, if this is needed or not.

@abulsayyad123 , @namrataukirde  any thoughts?

namrataukirde:
In case of appointment there are 2 scenario's :
- If the user chooses a purpose for organization and clicks on continue then it will take us good's detail's page. And then when we are on goods_details page and click on back button then it should take us to a request_purpose page.
- If a user chooses an option as for beneficiary then we take that user to client_information page. On click of continue, it will take us to goods_details page. On click of back button it should take us to client_information page and not to request_purpose page.

We can do it by checking the beneficiary record but when this feature was implemented we were not deleting beneficiary data if it is not needed.

Basically, that was done to handle backlinks. @patrixr, You can remove it if you have handled this scenario or used some other approach to handle it.

CC: @abulsayyad123 , @Darshan-Chauhan

abulsayyad123:
Yes @namrataukirde , we had used it to handle backlinks conditions. @patrixr , Yes we can remove it now. I had written this condition, but I don't think it is now required. 
cc: @Darshan-Chauhan

patrixr:
Thanks team. I think it's safe to remove. I'll do some proper testing as well

PR #313, opened 2019-02-25
GCW-2415 fix UI for browse messaging

abhinav098:
Hello Team, 
This PR fixes UI issues on browse app for messaging.
Thanks

### DIFF
--- a/app/components/variable-height-textarea.js
+++ b/app/components/variable-height-textarea.js
@@ -5,18 +5,28 @@ export default Ember.TextArea.extend({
   attributeBindings: ["disabled"],
   disabled: false,
 
+   didDestroyElement: function () {
+     Ember.$('.liquid-container').css({
+       "position": "relative",
+       "-webkit-transform": "translateY(0)",
+       "-moz-transform": "translateY(0)",
+       "transform": "translateY(0)"
+     });
+   },
+
   didInsertElement: function () {
+    Ember.$('.liquid-container').css({
+      "position": "unset",
+      "-webkit-transform": "unset",
+      "-moz-transform": "unset",
+      "transform": "unset"
+    });
     // scrolling down to bottom of page
     this.autoScroll();
   },
 
   autoScroll: function () {
-    let messageContainer = Ember.$(".message_container")[0];
-    if (messageContainer){
-      Ember.$(".message_container")[0].animate({
-        scrollTop: messageContainer.scrollHeight
-      }, 50);
-    }
+    window.scrollTo(0, document.body.scrollHeight);
   },
 
   handleReturnAndAutoScroll: function () {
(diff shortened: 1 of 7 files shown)


app/components/variable-height-textarea.js
```diff
@@ -5,18 +5,28 @@ export default Ember.TextArea.extend({
   attributeBindings: ["disabled"],
   disabled: false,
 
+   didDestroyElement: function () {
+     Ember.$('.liquid-container').css({
```
patrixr: is liquid container an ember thing ?

patrixr: `Ember.$('.liquid-container')` will fetch any liquid container across the whole app. Which sounds dangerous.
If this is really needed, make sure to only apply the changes on the liquid-container that matter (i'm guessing the one containing the textarea)

abhinav098: `liquid-container` is just a class defined to the main `div`, after body on browse. the CSS which I'm applying at the time of destroyComponent is already defined. All I'm doing is unsetting at the time of component render, and resetting it back as soon as the component goes away.

abhinav098: @patrixr i did not want to modify that thing globally so i took this approach of unsetting and resetting by using component lifecycle methods.

Darshan-Chauhan: Should we add a check if liquid container exists or not?

app/components/variable-height-textarea.js
```diff
@@ -5,18 +5,28 @@ export default Ember.TextArea.extend({
...
+       "position": "relative",
+       "-webkit-transform": "translateY(0)",
+       "-moz-transform": "translateY(0)",
+       "transform": "translateY(0)"
+     });
+   },
+
   didInsertElement: function () {
+    Ember.$('.liquid-container').css({
```
Darshan-Chauhan: ^

app/controllers/orders/conversation.js
```diff
@@ -20,9 +20,14 @@ export default detail.extend({
   sortedMessages: Ember.computed.sort("model.messages", "sortProperties"),
 
   groupedMessages: Ember.computed("sortedMessages", function () {
+    this.autoScroll();
     return this.groupBy(this.get("sortedMessages"), "createdDate");
   }),
 
+  autoScroll: function () {
```
Darshan-Chauhan: can we rename as `autoScroll()`?

app/components/variable-height-textarea.js
abhinav098: It's always there. But I'll add a check if you want.

app/controllers/orders/conversation.js
abhinav098: yeah sure.

namrataukirde:
:+1:

namrataukirde:
:+1:

PR #315, opened 2019-03-01
GCW-2414 read unread tracking of messages

abhinav098:
Hi team, this PR is for marking unread messages to read. 
Please review
Thanks

### DIFF
--- a/app/controllers/orders/detail.js
+++ b/app/controllers/orders/detail.js
@@ -6,22 +6,13 @@ import AjaxPromise from 'browse/utils/ajax-promise';
 export default applicationController.extend({
   order: Ember.computed.alias('model.order'),
   showCancelBookingPopUp: false,
-  orderSummaryTabs: [{
-      routeName: 'booking',
-      displayName: 'Booking',
-      icon: 'calendar'
-    },
-    {
-      routeName: 'goods',
-      displayName: 'Goods',
-      icon: 'shopping-basket'
-    },
-    {
-      routeName: 'conversation',
-      displayName: 'Messages',
-      icon: 'comment-o'
-    }
-  ],
+  unreadMessageCount: Ember.computed('order', function(){
+    return this.get('order.unreadMessagesCount');
+  }),
+  hasUnreadMessages: Ember.computed('order', function () {
+    return this.get('order.hasUnreadMessages');
+  }),
+
   deleteOrder(order) {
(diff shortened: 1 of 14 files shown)


app/controllers/subscription.js
```diff
@@ -201,6 +202,39 @@ export default Ember.Controller.extend({
         this.updateCartAvailability(0, cartItem);
       }
     }
+
+    // mark message read here
+    if (type === "message") {
```
patrixr: what's the alternative to `message` ? 🤔

app/controllers/subscription.js
```diff
@@ -201,6 +202,39 @@ export default Ember.Controller.extend({
...
+      var router = this.get("target");
+      var currentUrl = window.location.href.split("#").get("lastObject");
+
+      var messageRoute = this.get("messagesUtil").getRoute(data.item[type]);
```
patrixr: no real problem, but this method is really huge and complicated 😱 , if there's room for some refactoring into single-purpose methods that'd be a good improvement.

app/services/messages.js
```diff
@@ -0,0 +1,30 @@
+import Ember from "ember";
```
patrixr: Good call with a service, we need as much logic to be centralized 👍

patrixr: Actually, if there's other pieces of code that can be added to the service, that'd be a good way to go

app/controllers/orders/detail.js
```diff
@@ -6,22 +6,13 @@ import AjaxPromise from 'browse/utils/ajax-promise';
 export default applicationController.extend({
   order: Ember.computed.alias('model.order'),
   showCancelBookingPopUp: false,
-  orderSummaryTabs: [{
```
patrixr: isn't that needed for the tabs ? 😱

abhinav098: nope, not needed now, code has been changed.

app/services/messages.js
abhinav098: sure

app/controllers/subscription.js
abhinav098: there are many, order, item, version etc.

app/controllers/orders/detail.js
patrixr: cool :)

PR #316, opened 2019-03-04
GCW-2436 + Fix (some) cart merging bugs

patrixr:
Hi,

This PR addresses the following issues : 

- A cart merging issue due to a race-time condition (tries to merge before orders are loaded, #emberThings)
- Badly selected order due to computed property not being re-computed
- An issue where the server sometimes doesn't return related record, which would end up in the cart being faulty (added a check to manually make sure all packages are loaded)
- Using previous user's data due to failing `unloadAll` method (required update)
- Issue where backing from client_info back to request_purpose would re-create an order

Issues from GCW-2436 are related to the problems mentioned above

+ a bit of cleanup


app/controllers/request_purpose.js
```diff
@@ -63,14 +63,6 @@ export default Ember.Controller.extend(cancelOrder, {
       }
 
       let user = this.get('user');
-      let purposeIds = [];
```
patrixr: Quick mention regarding this block that got removed. The `1` and `2` are actually hardcoded DB Ids, which means this is basically a time bomb. The IDs on my machine had changed (re-ran the seeding) and this exploded.
Luckily, this logic had already been re-written in the updated client_information page, so I just removed this.

Let's keep our eyes out and make sure that doesn't occur anywhere else ! 😄 
The appropriate practice is to load the lookup table, and generate the options based on that.

ping @Darshan-Chauhan @abhinav098 @namrataukirde @abulsayyad123 @steveyken

namrataukirde: Completely agree with this @patrixr . We should not add hard-coded value anywhere in code.

app/controllers/request_purpose.js
```diff
@@ -87,7 +79,7 @@ export default Ember.Controller.extend(cancelOrder, {
       let orderParams = {
         organisation_id: user_organisation_id,
         purpose_description: this.get('description'),
-        purpose_ids: purposeIds,
+        purpose_ids: [],
```
namrataukirde: @patrixr As we have removed hard-coded values which are good. How we are sending purpose_ids to API?  Might Sound like a silly question but I think I am missing something.

patrixr: @namrataukirde , that is now done on the following page (client_information).
The radio buttons and logic have been moved there by señor @abulsayyad123  :D

--> https://github.com/crossroads/browse.goodcity/blob/master/app/controllers/order/client_information.js#L83

namrataukirde: That's Great. Thanks @patrixr , @abulsayyad123 :)

PR #322, opened 2019-03-06
GCW:2439 Sorted district selection

Darshan-Chauhan:
Hi Team,

This PR sorts districts alphabetically.
Jira ticket:- https://jira.crossroads.org.hk/browse/GCW-2439

Please review

### DIFF
--- a/app/controllers/request_purpose.js
+++ b/app/controllers/request_purpose.js
@@ -26,7 +26,7 @@ export default Ember.Controller.extend(cancelOrder, {
   selectedDistrict: null,
 
   districts: Ember.computed(function(){
-    return this.store.peekAll('district');
+    return this.store.peekAll('district').sortBy("name");
   }),
 
   getBookingTypeIdFor(identifier) {


abhinav098:
This looks great. We need to do the same changes for stock app also @Darshan-Chauhan.

patrixr:
> This looks great. We need to do the same changes for stock app also @Darshan-Chauhan.

would it maybe have been easier to do it on api side ?

abhinav098:
> would it maybe have been easier to do it on api side ?

Yes @patrixr that can also be done. Either we need to sort districts on the API or on the client. Simpler will be API, as we won't need to sort them on client side repeatedly. 

cc- @Darshan-Chauhan

Darshan-Chauhan:
@abhinav098 , there is issue with serializer if we sort district so I did on client side. I'll do it on stock too.

abhinav098:
👍

patrixr:
> @abhinav098 , there is issue with serializer if we sort district so I did on client side. I'll do it on stock too.

Those serializers are evil : )

steveyken:
I would generally leave sorting to the apps to do unless sending an `'?order=created_at` etc param but agree we have enough issues with serializers at the moment to warrant doing it solely client side. For small payloads like this, it'll be fine.

PR #328, opened 2019-03-11
Fix minor issue faced in book appointment

abhinav098:
Add query param in request purpose page, if its bookingtype and transitioning to request purpose.


namrataukirde:
Good Find 👍

abhinav098:
merging this 👍

PR #331, opened 2019-03-12
GCW-2422-Add online orders goods on confirm page

abulsayyad123:
Hi Team,
This PR is for JIRA ticket GCW-2422- Showing online orders on Confirmation page. I have added the online orders package block. Please review the PR and let me know if any change is required.
Thanks.

### DIFF
--- a/app/templates/order/confirm_booking.hbs
+++ b/app/templates/order/confirm_booking.hbs
@@ -35,7 +35,6 @@
         </div>
       </div>
     </section>
-
     {{!-- User details section --}}
     {{partial "order/user_details"}}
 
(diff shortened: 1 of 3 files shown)


app/templates/order/confirm_booking.hbs
```diff
@@ -52,6 +51,11 @@
       {{partial "order/goods_detail"}}
     {{/if}}
 
+    {{!-- Orders Package section --}}
+    {{#if order.isOnlineOrder }}
```
namrataukirde: What will happen in case of appointment? Do we want to show requests added on confirmation page?

abulsayyad123: It is already already there.

PR #337, opened 2019-03-27
GCW-2422 fix item details not visible

abhinav098:
This PR fixes the issues mentioned on GCW-2422 regarding items details invisible on confirmation screen. Please review

### DIFF
--- a/app/models/order.js
+++ b/app/models/order.js
@@ -47,16 +47,13 @@ export default Model.extend({
     return this.get("bookingType.isOnlineOrder");
   }),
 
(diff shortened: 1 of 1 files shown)


app/models/order.js
```diff
@@ -47,16 +47,13 @@ export default Model.extend({
     return this.get("bookingType.isOnlineOrder");
   }),
 
+  // only to show ordersPackages on confirmation screen
```
patrixr: the whole method can probably be replaced by
```
this.get("ordersPackages")
     .mapBy('record.package')
    .compact()
    .uniq()
```

PR #344, opened 2019-04-01
Fix minor issue when empty cart it should not go to account details if clicked on submit.

abhinav098:
Hello team is PR fixes a minor issue, where if there are no items in cart, on clicking submit button it should show the message box asking to add items to the cart, instead of redirecting to the account_details.
Thanks


namrataukirde:
:+1:

PR #346, opened 2019-04-03
GCW 2467 my dashboard

abhinav098:
Hi team this PR is for my dashboard screen
Thanks.

### DIFF
--- a/app/locales/en/translations.js
+++ b/app/locales/en/translations.js
@@ -62,7 +62,6 @@ export default {
       home: "Home",
       browse: "Browse",
       myorders: "My Orders",
-      myDashboard: "My Dashboard",
       faq: "Faq",
       about: "About",
       terms: "Terms",
@@ -272,7 +271,15 @@ export default {
   },
 
   my_dashboard: {
-    my_dashboard: "My Dashboard"
+    title: "My Dashboard",
+    updates: "Updates",
+    orders: "Orders",
+    account: {
+      account_text: "Account",
+      name: "Name",
+      account_details: "Account Details",
+      verified_text: "verified"
+    }
   },
 
   my_orders: {
--- a/app/locales/en/translations.js
+++ b/app/locales/en/translations.js
@@ -62,6 +62,7 @@ export default {
       home: "Home",
       browse: "Browse",
       myorders: "My Orders",
+      myDashboard: "My Dashboard",
       faq: "Faq",
       about: "About",
       terms: "Terms",
@@ -268,6 +269,10 @@ export default {
     cancel_booking_info:
       "This will cancel booking and all associated request for goods.",
     cancel_booking_reason_title: "Reason for canceling order"
+  },
+
+  my_dashboard: {
+    my_dashboard: "My Dashboard"
   },
 
   my_orders: {
--- a/app/locales/zh/translations.js
+++ b/app/locales/zh/translations.js
@@ -56,7 +56,6 @@ export default {
       home: "主頁",
       browse: "瀏覽",
       myorders: "我的訂單",
-      myDashboard: "My Dashboard",
       faq: "常問問題",
       about: "關於我們",
       terms: "條款",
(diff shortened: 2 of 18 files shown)


app/locales/en/translations.js
```diff
@@ -270,6 +271,10 @@ export default {
     cancel_booking_reason_title: "Reason for canceling order"
   },
 
+  my_dashboard: {
+    my_dashboard: "My Dashboard"
```
Darshan-Chauhan: Why do we need 2 translations for same text? i.e ** my_dashboard and myDashboard**

app/locales/zh/translations.js
```diff
@@ -56,6 +56,7 @@ export default {
       home: "主頁",
       browse: "瀏覽",
       myorders: "我的訂單",
+      myDashboard: "My Dashboard",
```
Darshan-Chauhan: I think above translation is not in this file?

app/templates/my_dashboard/_my_dashboard_tabs.hbs
```diff
@@ -0,0 +1,41 @@
...
+    {{#link-to 'my_account' }}
+      <i class="dashboard-menu">{{fa-icon 'vcard'}}</i>
+    {{/link-to}}
+  {{/link-to}}
+</div>
+
+<div class="row tabs tab-name active-tab">
+  <div class="small-4 columns updates tab text-center">
+  </div>
+  {{#link-to 'my_updates' tagName="div" classNames="small-4 updates tab active-tab columns text-center"}}
```
patrixr: how come there are 2 `link-to` within each other ?

app/routes/my_updates.js
```diff
@@ -0,0 +1,22 @@
+import AuthorizeRoute from "./authorize";
```
namrataukirde: Can we make parent controller for this? And just extend that? Code looks same at 2 places?

CC: @Darshan-Chauhan

abhinav098: sure @namrataukirde.

tests/acceptance/my-dashboard-test.js
```diff
@@ -0,0 +1,261 @@
...
+      assert.equal(Ember.$(".my_dashboard .title").text(), "My Dashboard");
+    });
+  });
+});
+
+test("should redirect to my_account page", function(assert) {
+  assert.expect(3);
+  visit("/");
+  andThen(function() {
+    click(Ember.$(".menu-3"));
```
namrataukirde: Instead of `.menu-3` can we add id for tab with tab name?

Just a thought.

app/locales/zh/translations.js
abhinav098: I think i need to check the translation file. I'll check and update. Sorry for the trouble guys.

tests/acceptance/my-dashboard-test.js
abhinav098: Yeah, we can but currently in the sidebar file, we have all menus like this only, will have to add classes to them also. 
Will make that change. Thanks for pointing out.

app/templates/my_dashboard/_my_dashboard_tabs.hbs
abhinav098: The above one is a `div` and lower one will become an anchor tag. 
cc - @namrataukirde

app/locales/en/translations.js
abhinav098: yeah, bymistake done, will remove the other one.

patrixr: it'd be pretty cool if we could reference translations, e.g `my_dashboard: "@some.other.key"`  ^^

app/templates/my_dashboard/_my_dashboard_tabs.hbs
patrixr: Can we just replace it with a `<div>` in that case, having 2 links embedded like that doesn't make much sense to me

app/templates/my_dashboard/_my_dashboard_tabs.hbs
```diff
@@ -0,0 +1,29 @@
+<div class="row tabs tab-icons">
+  {{#link-to 'my_updates' tagName="div" classNames="small-4 updates tab columns text-center"}}
```
patrixr: much better 😄

PR #349, opened 2019-04-08
Subscription refactor + Shallow load of orders

patrixr:
The following changes are made to allow the orders to live update
Part of that fix is also on the API side.

Notable changes :
 - refactor of subscription.js into a proper service. The custom logic that handles package changes was there was moved to `app/services/cart.js`
- removal of the preload of all orders for performance reasons, which led to making small changes in the places that failed to support shallow load


patrixr:
> This looks good. :) Awesome work.
> 
> Some test cases are failing though

Yeah, trying to fix them now. They're passing on my box though, very annoying 😆

PR #350, opened 2019-04-08
fix issue: mobile number not getting saved to db on email sign up

namrataukirde:
Hi Team,

This PR fixes issue: 
- If we sign up using email id and then enter mobile number on accounts details page then it do not saves it in db.
- PreferredContactNumber field was not considered while checking for account details are complete or not.

Please review.

Thanks.

### DIFF
--- a/app/controllers/account_details.js
+++ b/app/controllers/account_details.js
@@ -113,7 +113,7 @@ export default Ember.Controller.extend({
       user_attributes: {
         first_name: user.get("firstName"),
         last_name: user.get("lastName"),
-        mobile: this.mobileParam(),
+        mobile: this.mobileParam(user),
         email: user.get("email"),
         title: title
       }
@@ -125,10 +125,9 @@ export default Ember.Controller.extend({
     return params;
   },
 
-  mobileParam() {
-    var user = this.get("user");
-    var mobile = user.length && user.get("mobile");
-    if (mobile && mobile.length !== 0) {
+  mobileParam(user) {
+    var mobile = user && user.get("mobile");
+    if (mobile && mobile.length) {
       return mobile;
     } else {
       return "+852" + this.get("mobilePhone");
--- a/app/controllers/account_details.js
+++ b/app/controllers/account_details.js
@@ -12,11 +12,13 @@ export default Ember.Controller.extend({
   i18n: Ember.inject.service(),
   organisationId: Ember.computed.alias("model.organisation.id"),
   organisationsUserId: Ember.computed.alias("model.organisationsUser.id"),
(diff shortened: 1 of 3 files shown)


app/controllers/account_details.js
```diff
@@ -111,7 +113,7 @@ export default Ember.Controller.extend({
       user_attributes: {
         first_name: user.get("firstName"),
         last_name: user.get("lastName"),
-        mobile: user.get("mobile"),
+        mobile: this.mobileParam(),
```
Darshan-Chauhan: Can we pass user instead of getting user again in mobileParam?

app/controllers/account_details.js
```diff
@@ -123,6 +125,16 @@ export default Ember.Controller.extend({
     return params;
   },
 
+  mobileParam() {
+    var user = this.get("user");
+    var mobile = user.length && user.get("mobile");
+    if (mobile && mobile.length !== 0) {
```
Darshan-Chauhan:
```suggestion
    if (mobile && mobile.length) {
```

app/models/organisations_user.js
```diff
@@ -15,7 +15,25 @@ export default Model.extend({
...
+  hasPosition: Ember.computed("position", function() {
     return this.get("position") && this.get("position").length !== 0;
-  })
+  }),
+
+  hasPreferredContactNumber: Ember.computed(
+    "preferredContactNumber",
+    function() {
+      return (
+        this.get("preferredContactNumber") &&
```
Darshan-Chauhan: Can we cache `this.get("preferredContactNumber")?`

app/controllers/account_details.js
namrataukirde: Sure 👍

app/models/organisations_user.js
```diff
@@ -15,7 +15,25 @@ export default Model.extend({
     async: false
   }),
 
-  isInfoComplete: Ember.computed("position", function() {
+  hasPosition: Ember.computed("position", function() {
     return this.get("position") && this.get("position").length !== 0;
```
abulsayyad123: cache `this.get('position')`.

app/models/organisations_user.js
```diff
@@ -15,7 +15,25 @@ export default Model.extend({
...
-  })
+  }),
+
+  hasPreferredContactNumber: Ember.computed(
+    "preferredContactNumber",
+    function() {
+      return (
+        this.get("preferredContactNumber") &&
+        this.get("preferredContactNumber").length !== 0
```
abulsayyad123: cache `this.get('preferredContactNumber')`.

PR #354, opened 2019-04-10
Fetch packages when missing

patrixr:
When we come on the `goods` page from `my_orders` we don't have the packages already loaded (we don't load all dependencies at startup)

This code loads them if they are missing

### DIFF
--- a/app/controllers/application.js
+++ b/app/controllers/application.js
@@ -88,7 +88,9 @@ export default Ember.Controller.extend({
       this.get("cart").set("checkout", true);
       this.transitionToRoute("request_purpose", {
         queryParams: {
-          bookAppointment: false
+          onlineOrder: true,
+          bookAppointment: false,
+          orderId: null
         }
       });
     } else {
--- a/app/routes/orders/goods.js
+++ b/app/routes/orders/goods.js
@@ -1,11 +1,26 @@
 import detail from "./detail";
 
 export default detail.extend({
+  loadPackagesOf(order) {
+    return this.get("store")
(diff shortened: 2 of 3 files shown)


patrixr:
I've included the fix to https://jira.crossroads.org.hk/browse/GCW-2527 in this PR since it's really tiny

patrixr:
> I really wish that we didn't had to fire extra requests to fetch related data(request for each orders_package of order and each package of that orders_package, what would happen if orders_packages are 100s or 1000s :D ).
> Ember is relatively slow in rendering compare to react and this extra requests makes it more slower.
> If the app grows to enterprise size in future then we'll face lot's of issues.
> Sometimes I think that may we should re-architect ember structure for requests.

@Darshan-Chauhan 

All good points.

Few things regarding that : 
- I know it's a pain, but this "lazy" loading is probably the lesser of two evils. At least until we fix the underlying performance issues (serializers and indexes, @steveyken is working on it)
- I'm now implementing batch requests on Stock, I'll add those to Browse next, which should solve the large number of request issue
- Currently it **only** fetches things that it doesn't have and a single order with thousands of packages it pretty unlikely
- Re-architecturing is already something we've had in mind for a while (we experimented with GraphQL back in December).  However the opinionated architecture of Ember _expects_ things a certain way (cause you know... it's evil). So a server architecture would likely lead to a front-end re-arch as well (different frameworks). All things we will eventually get to

namrataukirde:
I agree with both of you @Darshan-Chauhan  and @patrixr. Good points and discussion. :+1:

PR #355, opened 2019-04-10
FIX => Goods Details not getting created 

abulsayyad123:
Hi Team,
This PR is for goods not getting added on goods_details page during Booking Appointment. Also fix for ticket GCW-2528-client details not getting updated (https://jira.crossroads.org.hk/browse/GCW-2528). Please, review this PR and let me know if any change is required.

Thanks.

### DIFF
--- a/app/controllers/order/goods_details.js
+++ b/app/controllers/order/goods_details.js
@@ -14,12 +14,9 @@ export default Ember.Controller.extend(cancelOrder, {
   qty: null,
   otherDetails: "",
   sortProperties: ["id"],
-  sortedGcRequests: Ember.computed(
+  sortedGcRequests: Ember.computed.sort(
     "order.goodcityRequests",
-    "sortProperties",
-    function() {
-      return this.get("order.goodcityRequests");
-    }
+    "sortProperties"
   ),
 
   hasNoGcRequests: Ember.computed("order.goodcityRequests", function() {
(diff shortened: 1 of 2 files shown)


app/controllers/order/goods_details.js
```diff
@@ -14,12 +14,16 @@ export default Ember.Controller.extend(cancelOrder, {
...
   otherDetails: "",
   sortProperties: ["id"],
-  sortedGcRequests: Ember.computed("gcRequests.[]", function() {
-    return this.get("gcRequests");
-  }),
+  sortedGcRequests: Ember.computed(
+    "order.goodcityRequests",
+    "sortProperties",
+    function() {
+      return this.get("order.goodcityRequests");
```
Darshan-Chauhan: Is this sorted?? Because it doesn't look like it is

abulsayyad123: Yes, it is sorted according to order of its creation.

Darshan-Chauhan: Because I see it's dependent on `sortProperties` but it is not used

abulsayyad123: Got the issue, thanks for clarifying. Added sortProperties in sort computed property.

app/models/order.js
```diff
@@ -97,15 +97,15 @@ export default Model.extend({
     }
   ),
 
-  clientIdNumber: Ember.computed("beneficiary", function() {
+  clientIdNumber: Ember.computed("beneficiary.identityNumber", function() {
```
patrixr: those can probably be shortened with `Ember.computed.alias('beneficiary.identityNumber')` :D

PR #357, opened 2019-04-16
GCW-2516 Redesign home page and add some static pages

abhinav098:
Hello team, 
This PR adds static pages, for about, faqs, terms and privacy on browse.
Also this PR makes changes to the home screen of browse, to look somewhat like www.goodcity.hk
Please review this PR.
Thanks.


app/routes/static_pages.js
```diff
@@ -0,0 +1,24 @@
+import accountDetails from "./account_details";
+
+export default accountDetails.extend({
+  setupController(controller, model, transition) {
+    this._super(...arguments);
+    let applicationController = this.controllerFor("application");
+    applicationController.set("showSidebar", true);
+    let pageTitle =
```
steveyken: this doesn't feel quite right. is there a way to pass in the pageTitle or use the targetName inside the i18n get...

```
this.get("i18n").t(transition.targetName + ".title")
```

Is that safe? 
cc @patrixr

abhinav098: yeah, sure, I'll try this out, and change it. Its even better 👍 
Thanks @steveyken .

patrixr:
```
this.get("i18n").t(`${transition.targetName}.title`)
```

I think it's fine to rely on translation keys, as long as it works seamlessly when we add new sections in the translation file. Doing `if/else` means that every time we add something in the file we need to come back here.

PR #361, opened 2019-04-23
GCW-2516 fix UI related issues for large screens

abhinav098:
Hello Team, 
This PR fixes the issues for the UI layout on larger screens.
Thanks

### DIFF
--- a/app/styles/_sidebar.scss
+++ b/app/styles/_sidebar.scss
@@ -119,7 +119,7 @@ $sidebar-width: 7.8rem;
 .off-canvas-wrap {
   @media only screen and (min-width: 71.063em) and (max-width: 120em){
     margin-left: 4rem;
-    width: 94.5%;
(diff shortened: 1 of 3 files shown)


app/styles/_sidebar.scss
```diff
@@ -175,7 +175,7 @@ $sidebar-width: 7.8rem;
     -ms-transform: none;
     -o-transform: none;
     transform: none;
-    margin-left: $sidebar-width;
+    margin-left: 5.8rem;
```
Darshan-Chauhan: What was the value of sidebar-width? 
Will this affect on size of sidebar or how it's shown in UI?

abhinav098: sidebar-width is 7.8em, 
I have changed it only here, not on the global variable so it will not affect the size of sidebar.

PR #369, opened 2019-04-30
FIX=> Client information fix not appearing on online order creation

abulsayyad123:
Hi Team,
This PR is fix for Client information not appearing on confirm booking page on first time creation.  Please review the PR and let me know if any change is required.
Thanks

### DIFF
--- a/app/controllers/order/client_information.js
+++ b/app/controllers/order/client_information.js
@@ -3,6 +3,7 @@ import config from "../../config/environment";
 import AjaxPromise from "browse/utils/ajax-promise";
 const { getOwner } = Ember;
 import cancelOrder from "../../mixins/cancel_order";
+import { task } from "ember-concurrency";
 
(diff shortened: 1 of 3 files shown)


app/controllers/order/client_information.js
```diff
@@ -47,102 +50,124 @@ export default Ember.Controller.extend(cancelOrder, {
...
     } else if (!isOrganisation && !beneficiaryId) {
-      actionType = 'POST';
+      actionType = "POST";
     } else {
       actionType = null;
     }
     return actionType;
   },
 
+  editOrder: task(function*(orderParams, isOrganisation) {
```
Darshan-Chauhan: Why use task and not async/await?

abulsayyad123: I had used async/await but tests are failing because of code coverage plugin.

PR #377, opened 2019-05-10
FIX=> Pushing sidebar below download banner

abulsayyad123:
Hi Team,
This is fix for pushing sidebar menu below download banner. Please review the PR and let me know if any change is required. 
Thanks

### DIFF
--- a/app/controllers/application.js
+++ b/app/controllers/application.js
@@ -131,6 +131,10 @@ export default Ember.Controller.extend({
   },
 
   actions: {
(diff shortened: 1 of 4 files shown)


app/controllers/application.js
```diff
@@ -131,6 +131,10 @@ export default Ember.Controller.extend({
   },
 
   actions: {
+    removeSidebarMargin() {
```
abhinav098: can we rename this method to something else ?

PR #382, opened 2019-05-18
fix account details shifting to right 

abhinav098:
On clicking book appointment button from home page, content of account details shifts to right. 
This PR contains its fix.
Please review 
Thanks


namrataukirde:
:+1:

PR #385, opened 2019-05-20
GCW-2586-Organisation selection issue from Account Details page

abulsayyad123:
Hi Team,
This PR has changes related button and on clicking it redirects user back to account details form. And some minor CSS fix added. Please review the PR and let me know if any change is required. 
Thanks

### DIFF
--- a/app/routes/search_organisation.js
+++ b/app/routes/search_organisation.js
@@ -1,4 +1,7 @@
-import AuthorizeRoute from './authorize';
+import AuthorizeRoute from "./authorize";
 
-export default AuthorizeRoute.extend({});
-
+export default AuthorizeRoute.extend({
+  setupController() {
+    this.controllerFor("application").set("showSidebar", false);
+  }
+});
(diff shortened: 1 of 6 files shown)


app/routes/search_organisation.js
```diff
@@ -1,4 +1,7 @@
-import AuthorizeRoute from './authorize';
-
-export default AuthorizeRoute.extend({});
+import AuthorizeRoute from "./authorize";
 
+export default AuthorizeRoute.extend({
+  setupController() {
+    this.controllerFor("application").set("showSidebar", false);
+  }
```
abhinav098: do we need a reset controller for this ? @abulsayyad123 . I think we do.

abulsayyad123: either of them works.

abhinav098: No, i'm saying we need to revert this property to whatever it was, while controller is exiting. 
Because its application level property. It might affect other places.

abulsayyad123: I am sorry, I miss understood. Yes, you are right. I intentionally did not reset it because from this page we redirect to `account_details` page in all possible cases. And in that page it is already doing this reset.

abhinav098: okkay, but ideally we should do it. What if someone navigates through the url to lets say home ?
sidebar won't be visible. That'll cause issue over there.

PR #396, opened 2019-05-24
FIX=> GCW-2595- Confirmation page not getting updated on adding items to cart

abulsayyad123:
Hi Team,
This is fix for confirmation page not getting updated on adding items to cart. Please review the PR and let me know if any change is required. 

Thanks

### DIFF
--- a/app/controllers/request_purpose.js
+++ b/app/controllers/request_purpose.js
@@ -49,7 +49,7 @@ export default Ember.Controller.extend(cancelOrder, {
 
   isOnlineOrder() {
     return (
-      this.getSelectedBookingTypeId() ===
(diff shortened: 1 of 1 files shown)


app/controllers/request_purpose.js
```diff
@@ -49,7 +49,7 @@ export default Ember.Controller.extend(cancelOrder, {
 
   isOnlineOrder() {
     return (
-      this.getSelectedBookingTypeId() ===
+      this.getSelectedBookingTypeId() ==
```
Darshan-Chauhan: Why type check was removed? I'd prefer to convert string to integer rather than removing type check. 
This is fine but for sake of consistency use `===` and convert string to integer using `+`;

PR #402, opened 2019-06-04
Fix broken sidebar (whoopsie)

patrixr:
Looks like my previous fix was too hopeful, they hbs and jquery are too intertwined it's hard not to break things :/


abhinav098:
👍

PR #404, opened 2019-06-10
GCW 2616 browse goods search

abhinav098:
Hello @team,
This PR adds the ability to search items on browse app.
Please review this PR.
Thanks.

### DIFF
--- a/app/controllers/search_goods.js
+++ b/app/controllers/search_goods.js
@@ -73,7 +73,7 @@ export default BrowseController.extend({
       const params = this.trimQuery(
         _.merge({}, this.getSearchQuery(), this.getPaginationQuery(pageNo))
       );
-      return this.get("store").query("package", params);
+      return this.store.query("package", params);
     }
   }
 });
--- a/app/controllers/search_goods.js
+++ b/app/controllers/search_goods.js
@@ -1,7 +1,7 @@
-import Ember from "ember";
+import BrowseController from "./browse";
 import _ from "lodash";
 
-export default Ember.Controller.extend({
+export default BrowseController.extend({
   minSearchTextLength: 2,
   filteredResults: [],
   displayResults: false,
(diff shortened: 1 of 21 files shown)


app/controllers/search_goods.js
```diff
@@ -0,0 +1,92 @@
+import Ember from "ember";
+import searchModule from "./search_module";
+import _ from "lodash";
+
+export default searchModule.extend({
```
patrixr: I'm assuming you're using @abulsayyad123 's infinite-list

If that's the case, there's no need to inherit from `search_module`. We also want to avoid that as much as possible (we want to strip out that nightmare class, and eventually delete it altogether).

app/controllers/search_goods.js
```diff
@@ -0,0 +1,92 @@
...
+    });
+  },
+
+  showResults() {
+    Ember.run(() => {
+      this.set("displayResults", true);
+    });
+  },
+
+  getFilterQuery() {
```
patrixr: Clean class 👍 
I like the dividing the queries in those functions, even if some are static content and could technically all be grouped, it shows me exactly where to edit them and what they are

app/controllers/search_goods.js
```diff
@@ -0,0 +1,92 @@
...
+    return {
+      state: "in_stock,published"
+    };
+  },
+
+  getSearchQuery() {
+    return {
+      searchText: this.get("searchText"),
+      browseSearch: true
```
patrixr: @abhinav098  could you share what that property does and why we need it ?

patrixr: never mind, saw the api pull request and added a comment there : )

app/controllers/search_goods.js
abhinav098: yeah sure, inherited it from ember component now.

app/controllers/search_goods.js
abhinav098: yeah right

app/controllers/search_goods.js
abhinav098: yep, understood, and removed the property

abhinav098:
Merging this branch, as child branch has approving reviews.

PR #405, opened 2019-06-11
Gcw 2617 fix reported issues

abhinav098:
This PR fixes the issues reported by Tim on GCW-2617.
Please review thanks.

### DIFF
--- a/app/controllers/item.js
+++ b/app/controllers/item.js
@@ -1,76 +1,91 @@
-import Ember from 'ember';
+import Ember from "ember";
 import config from "../config/environment";
 
 export default Ember.Controller.extend({
-
-  messageBox:     Ember.inject.service(),
-  application:    Ember.inject.controller(),
-  queryParams:    ['categoryId', 'sortBy'],
-  categoryId:     null,
-  cart:           Ember.inject.service(),
-  sortBy:         "createdAt",
-  item:           Ember.computed.alias('model'),
-  noNextItem:     Ember.computed.empty('nextItem'),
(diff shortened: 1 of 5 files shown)


app/controllers/item.js
```diff
@@ -1,76 +1,91 @@
...
+    "item.packages.@each.orderId",
+    "item.isUnavailableAndDesignated",
+    function() {
+      var currentPath = this.get("target").currentPath;
+      var item = this.get("item");
+      var isItemUnavailable = this.get("item.isUnavailableAndDesignated");
+      if (
+        (currentPath === "item" || currentPath === "package_category") &&
+        isItemUnavailable &&
+        isItemUnavailable !== null &&
```
patrixr: `isItemUnavailable && isItemUnavailable !== null` can be replaced with `isItemUnavailable` 

I think all those changes are from prettier, but it caught my eye

app/controllers/item.js
```diff
@@ -1,76 +1,91 @@
...
+        !this.get("itemNotAvailableShown")
+      ) {
+        this.set("itemNotAvailableShown", true);
+        if (this.get("cart").hasCartItem(item)) {
```
patrixr: A boolean flag (here `hasQuantityAndIsAvailable `) should never modify anything, that's a typical [SIDE EFFECT](https://en.wikipedia.org/wiki/Side_effect_(computer_science))

no need to modify it, I'm rewriting the cart anyways : )

abhinav098: yep all these are from prettier

PR #412, opened 2019-06-25
Cloud-based cart

patrixr:
Hi team,

Here is the rewrite of the cart functionality, as well as a lot lot lot of refactoring and cleanup.
Brace yourselves :D

## The process

- A guest user adds items/packages to the cart, this create local `cart_item` records 
- When the user logs in, we loop over those records and run `.save()` , which persists them in the cloud
- When a logged in user requests something, it just creates a record and saves it
- When the user submits an order (at the **end** of the order process), it calls the `/checkout` endpoint, which empties the cart and returns the submitted order with all it's orders_packages

## Notes

The cart only contains packages, they appear as items only if : 
- all sibling packages are in the cart
- all sibling packages are available

## Links to main changes 

Here are the biggest, most critical changes (to help in the review) :

- Complete redo of the cart service 
https://github.com/crossroads/browse.goodcity/pull/412/files#diff-18037e49dab72fe695af8883871123f6R11

- The "submit" process has now changed a bit (calling /checkout)
https://github.com/crossroads/browse.goodcity/pull/412/files#diff-ac00f01f5ad265ad158e6494df33977dR13

- New cart_item model
https://github.com/crossroads/browse.goodcity/pull/412/files#diff-2b678f993ba720207be2b7c80ceea5aeR6

- Moved a lot of functionalities into services, to avoid having that repeated AjaxPromise everywhere
https://github.com/crossroads/browse.goodcity/pull/412/files#diff-1f8472d40d6b4b760487eed59a58c2a0R3

- The "preload_mixin" is not a service, it should never have been a mixin in the first place imo, as it's just a helper function. https://github.com/crossroads/browse.goodcity/pull/412/files#diff-de4edd8f5f50dd9bb1d75d806fa651deR11

- The rest is a LOT of cleanup, so many things were re-implemented everywhere, so I moved functionalities into services


app/components/book-appointment.js
```diff
@@ -1,28 +1,10 @@
...
-    return (
-      hasInfoAndCharityRole && organisation && hasCompleteOrganisationUserInfo
-    );
-  },
-
+  session: Ember.inject.service(),
   actions: {
     redirectAsPerUserDetails() {
-      if (this.isDetailsComplete()) {
+      if (this.get("session").accountDetailsComplete()) {
```
steveyken: Great! Abstract behind a service and re-use elsewhere

app/controllers/order/confirm_booking.js
```diff
@@ -1,52 +1,52 @@
...
-    return true;
+  submitOrder(order) {
+    return this.runTask(
+      order.get("isOnlineOrder")
+        ? this.get("cart").checkoutOrder(order)
+        : this.get("orderService").submitOrder(order)
+    );
+  },
+
+  badCart() {
```
steveyken: watch out for cousin 'naughty_cart'

app/instance-initializers/ajax.js
```diff
@@ -0,0 +1,13 @@
+import AjaxPromise from "browse/utils/ajax-promise";
```
steveyken: nice!

app/locales/en/translations.js
```diff
@@ -28,7 +28,7 @@ export default {
   awaiting_dispatch: "awaiting dispatch",
   dispatching: "dispatching",
   items_not_available:
-    "One or many items in your cart are no longer available. Please remove those items from cart and then proceed.",
+    "One or many items in your cart are no longer available.",
```
steveyken:
```suggestion
    "Some of the items in your cart are no longer available.",
```

app/mixins/async_tasks.js
```diff
@@ -0,0 +1,61 @@
...
+        return res;
+      })
+      .catch(err => {
+        this.__incrementTaskCount(-1);
+        return Ember.RSVP.reject(err);
+      });
+  },
+
+  showLoadingSpinner() {
+    if (Ember.testing) {
```
steveyken: Does this mess up tests? Any way to work around it so the code does run during tests?

app/mixins/cancel_order.js
```diff
@@ -1,51 +1,59 @@
...
+    if (!cancellationReason) {
+      div.addClass("cancel-booking-error");
       return false;
-    } else {
-      Ember.$(`#cancel-appointment-reason${order.id}`).removeClass("cancel-booking-error");
-      this.set("cancellationReasonWarning", false);
     }
-    var url = `/orders/${order.id}/transition`;
-    var loadingView = getOwner(this).lookup('component:loading').append();
-    new AjaxPromise(url, "PUT", this.get('session.authToken'), { transition: "cancel", cancellation_reason: cancellationReason })
```
steveyken: great to see this abstracted out

app/services/api-base-service.js
```diff
@@ -0,0 +1,69 @@
+import Ember from "ember";
+import AjaxPromise from "browse/utils/ajax-promise";
+
+export default Ember.Service.extend({
```
steveyken: 💯

app/services/local-storage.js
```diff
@@ -0,0 +1,48 @@
...
+    return obj && obj.data;
+  } catch (e) {
+    console.error(e);
+    return null;
+  }
+}
+
+function prefixKey(key) {
+  if (isProd) {
+    return key;
```
steveyken: Any reason why we don't prefix in production as well?

patrixr: I never prefixed it because if someone updates their app, it suddenly won't find the keys anymore. The prefix is mostly to prevent test data to be present when we run the app

app/mixins/async_tasks.js
patrixr: Yeah, there was a reeeaaalllly dodgy failing test once.  I never did figure out another way to fix it, It had something to do with the loading spinner being removed at odd moments.

steveyken: Shame but understandable!

app/services/cart.js
```diff
@@ -1,179 +1,382 @@
...
+    return this.addPackage(pkgOrItem);
+  },
+
+  /**
+   * Adds the packages of the item to the cloud cart.
+   * If the user is not logged in, will only save it locally
+   */
+  addItem(item) {
+    return Ember.RSVP.all(
+      item.get("packages").map(pkg => this.addPackage(pkg))
```
abulsayyad123: is it fine to use `all` here ? I think it should be used for multiple promises here it is one promise having multiple promises. Rectify me if I am wrong.

patrixr: `map` returns an array, so it is multiple promises. one for each package :)

app/services/cart.js
```diff
@@ -1,179 +1,382 @@
...
-  cartItems: computed("[]", "@each.available", function() {
-    var content = this.get("content");
-    var allItems = [];
-    content.forEach(record => {
-      if (record.available) {
-        allItems.push(
-          this.get("store").peekRecord(record.modelType, record.id)
-        );
-      }
+    this.clearDuplicates();
```
abulsayyad123: What is the reason to write this method `clearDuplicates` ? Why this Duplicates are getting added 🤔

app/components/book-appointment.js
abulsayyad123: 👍

app/services/cart.js
patrixr: The case is : 

1. You log-in, and add items to your cart
2. You start an new session as a guest user (other device for example)
3. You add an item, as a guest, that is already in your cart
4. You log-in, the item is both in your cart, and stored locally as a "guest item"
5. We do a check to make sure that it's not there twice

The server would not allow the package to exist twice in the cart. But rather than having a pop-up error message I thought it would be best to clear those that are already added in the cart

app/mixins/async_tasks.js
```diff
@@ -0,0 +1,61 @@
+import Ember from "ember";
+const { getOwner } = Ember;
+
+export default Ember.Mixin.create({
```
abulsayyad123: Excellent !! work here. Do we need to use this mixin with other controllers to avoid use of spinners in future ?

app/services/api-base-service.js
abulsayyad123: 👍

app/mixins/async_tasks.js
patrixr: You don't "need" to, but it's there as a helper to make the controllers leaner and smaller

patrixr: I don't want to enforce a practice, so I try it out, and see if it sticks : )
But I definitely don't like the way it's currently done (copy pasted every time)

app/services/cart.js
abulsayyad123: Is there any utility in lodash, which avoids duplicates while merging offline items with existing API cart ? Just a thought

app/components/book-appointment.js
namrataukirde: 👍

app/services/api-base-service.js
namrataukirde: 👌

PR #423, opened 2019-07-09
change the backlink name on item page to "Back to Search"

abhinav098:
This PR fixes one issue which TIM reported, that on clicking item from search page, the back link appears as category name, but it takes back to search page. 
So changes have been made such that, when user comes from search_page, it'll show "Back to search" otherwise it will show category name as the back link.
Also, this PR contains some css fixes for search page. Please review. 
Thanks

### DIFF
--- a/.ember-cli
+++ b/.ember-cli
@@ -5,5 +5,6 @@
 
     Setting `disableAnalytics` to true will prevent any data from being sent.
   */
-  "disableAnalytics": false
+  "disableAnalytics": false,
+  "port": 4202
 }
--- a/app/controllers/item.js
+++ b/app/controllers/item.js
@@ -5,6 +5,7 @@ export default Ember.Controller.extend({
   messageBox: Ember.inject.service(),
   application: Ember.inject.controller(),
   queryParams: ["categoryId", "sortBy"],
+  prevPath: null,
   categoryId: null,
   cart: Ember.inject.service(),
   sortBy: "createdAt",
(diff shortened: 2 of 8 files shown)


.ember-cli
```diff
@@ -5,5 +5,6 @@
 
     Setting `disableAnalytics` to true will prevent any data from being sent.
   */
-  "disableAnalytics": false
+  "disableAnalytics": false,
+  "port": 4202
```
abhinav098: This has been added to run the app using "ember serve" earlier "ember serve" used to start the server on 4200, not it will start on 4202 for browse app.

app/controllers/item.js
```diff
@@ -72,6 +73,14 @@ export default Ember.Controller.extend({
     return this.store.peekRecord("package_category", this.get("categoryId"));
   }),
 
+  linkDisplayName: Ember.computed("prevPath", function() {
```
patrixr: could we move that to the template file ?
That way we keep data and rendering separate :)

```
{{#if (is-equal prevPath "search_goods")}}
     {{t 'search_goods.back'}}
{{else}}
    {{ categoryObj.name }}
{{/if}
```

abhinav098: Yeah, it can be done, but i have purposely done that to separate out logic from the view files to the controller.
Let me know, if i should change it !

app/controllers/item.js
```diff
@@ -72,6 +73,14 @@ export default Ember.Controller.extend({
...
+    let prevPath = this.get("prevPath");
```
Darshan-Chauhan: Should start using `const`

PR #424, opened 2019-07-09
GCW-2617 breadcrumbs improvements

abhinav098:
This PR contains some improvements in the Breadcrumbs selection process. 
Please review and comment.


app/templates/browse.hbs
```diff
@@ -33,9 +33,9 @@
             </div>
             <ul class="list subcategories hide-for-small">
               {{#each category.childCategories as |childCategory|}}
-                {{#link-to 'package_category' childCategory.id class="custom_a"}}
+                <div class="custom_a" {{action "viewChildCategoryItems" childCategory}}>
```
abulsayyad123: You can try `link-to` tag along with action to remove `viewChildCategoryItems`.
```suggestion
     {{#link-to 'package_category' childCategory.id class="custom_a"}}
        <div {{action 'setChildCategory' target="packageCategory">
```
And inside `packageCategory` controller add one method `setChildCategory` which will have this set action.
```
 actions: {
    setChildCategory(category) {
       this.set("selectedCategoryId", category)
   }
 }
```

app/controllers/package_category.js
```diff
@@ -31,12 +33,18 @@ export default Ember.Controller.extend({
     ];
   }),
 
+  isCategorySelected: Ember.computed("selectedCategoryId", function() {
+    return !!this.get("selectedCategoryId.id");
```
abulsayyad123: Reason for this `!!` ?

app/controllers/package_category.js
```diff
@@ -31,12 +33,18 @@ export default Ember.Controller.extend({
...
+  }),
+
   categoryObj: Ember.computed("selectedCategoryId", "model", function() {
     this.set("page", 1);
     var selectedCategoryId = this.get("selectedCategoryId.id");
-    return selectedCategoryId
-      ? this.store.peekRecord("package_category", selectedCategoryId)
-      : this.get("model");
+    if (selectedCategoryId) {
```
abulsayyad123: can we use ternary operator here ?

app/controllers/package_category.js
abhinav098: Screen Shot 2019-07-10 at 5 26 58 PM
converting it to boolean.

abulsayyad123: You can use bool computed property for this.
https://api.emberjs.com/ember/2.15/namespaces/Ember.computed/methods?anchor=bool

app/controllers/browse.js
```diff
@@ -3,6 +3,7 @@ import config from "../config/environment";
 
 export default Ember.Controller.extend({
   isMobileApp: config.cordova.enabled,
+  packageCategory: Ember.inject.controller(),
```
patrixr: How come we need the packageCategory controller ?
I feel like injecting a controller into another is generally a code smell, but I also know that we're sometimes forced to (been guilty of that myself).
So just want to know the reason, maybe we can come up with something better 😄

app/controllers/package_category.js
patrixr: I guess the bool property would be the more "ember appropriate" way of doing it. Although `!!` kinda makes us cooler. Tough call 😄 (either way is fine by me)

app/templates/_navbar.hbs
```diff
@@ -0,0 +1,20 @@
+<nav class="breadcrumbs">
+  {{!-- all goods --}}
```
patrixr: I like the comments

app/templates/package_category.hbs
```diff
@@ -7,12 +7,9 @@
...
         {{/link-to}}
       </section>
-      <nav class="breadcrumbs">
-        {{#link-to 'browse' classNames="underline-link"}}{{t 'browse.all_goods'}}{{/link-to}}
-        {{#link-to 'package_category' model.id classNames="current"}}{{model.name}}{{/link-to}}
-        {{select-list class="category_select" content=selectCategories on-change=(action (mut selectedCategoryId)) prompt=(t 'browse.filter_results')}}
-        <i class="fa fa-caret-down category_select"></i>
-      </nav>
+      <div class="browse-item-navbar">
+        {{partial 'navbar'}}
```
patrixr: @abhinav098 do you think we can rename `navbar` to maybe something like `breadcrumbs` ?
I know I'll confuse `navbar` the bar at the top of the app, or the sidebar. You need "patrick-proof" names 😆

app/controllers/item.js
```diff
@@ -127,6 +128,15 @@ export default Ember.Controller.extend({
       );
     },
 
+    viewChildCategory(category) {
```
patrixr: shouldn't this be `viewParentCategory` ? since it's navigating to `category.get("parentId")` ? 🤔 

(maybe that function could be shared with the browse controller somehow, they look pretty similar btw)

app/templates/browse.hbs
patrixr: would both of them trigger ? 🤔 

I think `setChildCategory` can also be removed by using the `mut` helper 
https://medium.com/@_edhuang/ember-mut-helper-lets-mutate-acdf6b58a0df

I've never used it though, so not 100% sure. I think @Darshan-Chauhan might know

app/controllers/browse.js
abhinav098: I want to set a property of packageCategory from this controller.

app/templates/browse.hbs
abhinav098: Thanks for the suggestion @abulsayyad123 , but this wont work. Will break it. and the functionality which is required will not work with this code.

abhinav098: `{{#link-to 'package_category' childCategory.id}}`
This will generate url as `/category/{{childCategoryId}}` which is not the requirement. We should be able to switch between categories from the parent url without changing the URL.
Thanks

abhinav098: I can change the method name if you say so, Will rename it to `setChildCategory`.
Do let me know in case.
Thanks @abulsayyad123

app/controllers/browse.js
abhinav098: maybe a service can be created for this, and this code can be moved to the service to avoid controller injection. Do let me know your thoughts on this @patrixr .

app/templates/package_category.hbs
abhinav098: sure, I'll rename it to `_breadcrumbs.hbs`

app/controllers/item.js
abhinav098: yeah, it is navigating to parent category but, here child category is preselected. on clicking any sub category from `browse` page @patrixr .
Maybe we can rename it to `setChildCategoryAndRedirect`
Let me know if it sounds good !!

app/templates/browse.hbs
abulsayyad123: using `eventpropagation` it can be avoided. My suggestion was considering repetition of same logic twice in controller. So, i tried it to handle it within templates itself.

PR #426, opened 2019-07-11
Cart waits for user profile to be loaded

patrixr:
In order to save the packages that were selected while offline, we need the user ID.

But the current logic does `if (authToken) { ... }` , however the authToken doesn't guarantee that the user profile was downloaded yet. This aims to fix that.

Also, since I got the chance. I also changed the way the current user info is retrieved from the store.

We currently had the currentUser equal to `this.get("store").peekAll("user").get("firstObject")` . It's obviously flaky and wrong to assume that the first user in the data is the logged in user.
So I've made it that the current user is dependant on the ID that we fetch when we load the profile !

### DIFF
--- a/app/services/cart.js
+++ b/app/services/cart.js
@@ -16,7 +16,10 @@ export default ApiService.extend({
 
   init() {
     this._super(...arguments);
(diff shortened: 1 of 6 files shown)


namrataukirde:
👌

PR #432, opened 2019-07-15
Quick fix for the account details page navigation bug

patrixr:
Here's the fix for one of the `account_details` page.

Note: I think the whole navigation flow needs some work, but making a "hotfix", since we're a bit short on time before the launch event.

## The problem

A lot of this is hard to see when reading to code, because we need to follow the transitions all the way, but here's a breakdown of what was happening:

For a new user, the current logging flow looked like this : 

`[auth page] -> [verification page] -> [account details page]`

And the "submit cart" flow as a `guest` looked like this: 

`[login flow] -> [account details page] -> [request_purpose]`

So If I write the whole thing it essentially resulted in this :

`[auth page] -> [verification page] -> [account details page] -> [account details page] -> [request_purpose]`

The `account_details` page would be there twice, and when someone would confirm the details for the second time, the server answers with "email already exists" (or such)

Some people would also not click a second time, and see the accoun_details page as "stuck"

### DIFF
--- a/app/models/package.js
+++ b/app/models/package.js
@@ -42,12 +42,12 @@ export default Model.extend(cloudinaryImage, {
   isDispatched: Ember.computed.bool("stockitSentOn"),
 
   isAvailable: Ember.computed("isDispatched", "allowWebPublish", function() {
-    return (
+    return Boolean(
       !this.get("isDispatched") &&
-      (this.get("allowWebPublish") ||
-        (this._internalModel._data &&
-          this._internalModel._data.allowWebPublish)) &&
-      this.get("quantity")
+        (this.get("allowWebPublish") ||
+          (this._internalModel._data &&
+            this._internalModel._data.allowWebPublish)) &&
+        this.get("quantity")
     );
   }),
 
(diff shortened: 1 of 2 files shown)


app/models/package.js
```diff
@@ -42,12 +42,12 @@ export default Model.extend(cloudinaryImage, {
   isDispatched: Ember.computed.bool("stockitSentOn"),
 
   isAvailable: Ember.computed("isDispatched", "allowWebPublish", function() {
-    return (
+    return Boolean(
```
abulsayyad123: Before it wasn't returning boolean ?

abhinav098: Same question @patrixr , What was it returning before(if not boolean)?

patrixr: Javascript has the concept of things being "truthy", which allows us to do things like this :

```javascript
const str = "a string";
if (str) {
   ...
}
```

But the result of the expression is still not a boolean.
In the case of an `&&` expression, the value will equal the first "falsy" result. If all of them are "truthy" then the last value is used

Meaning : 

```
1 && 2 && 0 && 4  => 0
1 && 2 && "hello" => "hello"
"hello" && 2 && 1 => 1
```

Here the value would be `this.get("quantity")` which is an integer.
Which would fail later when using code suche as `.filterBy('isAvailable', true)` (because 1 !== true)

patrixr: ❤️ All hail Typescript ❤️

abhinav098: great explanation @patrixr thanks :) 👍

patrixr: 😄

PR #433, opened 2019-07-15
GCW-2683-Allow to add items to submitted order

abulsayyad123:
Hi Team,
This PR is for JIRA ticket GCW-2683- Allow to add items to submitted orders. Please review this PR and let me know if any change is required.

**Note**: On `goods.hbs` page there is some issue with indentation which is not reflecting in my editor.


app/routes/orders/goods.js
```diff
@@ -1,6 +1,16 @@
 import detail from "./detail";
 
 export default detail.extend({
+  showMessage: false,
+
+  beforeModel(transition) {
+    const previousRoutes = this.router.router.currentHandlerInfos;
+    const previousRoute = previousRoutes && previousRoutes.pop().name;
+    if (previousRoute === "submitted_orders") {
```
patrixr: I'd suggest that the message is passed to the page, rather than having the page check the history.
That makes it more modular.

app/routes/submitted_orders.js
```diff
@@ -0,0 +1,54 @@
...
+  async beforeModel() {
+    const hasSubmittedOrders = await this.get(
+      "orderService"
+    ).hasSubmittedOrders();
+
+    if (!hasSubmittedOrders) {
+      let lastDraftOrder = await this.get("orderService").getLastDraft({
+        appointment: false
+      });
+      const orderId =
```
patrixr: `lastDraftOrder` is a single model, so it should not have a `length`
I think a `const orderId = lastDraftOrder && lastDraftOrder.get('id')` works here

app/services/order-service.js
```diff
@@ -118,12 +118,34 @@ export default ApiService.extend({
...
+      orders => {
+        return orders
+          .filterBy("state", "draft")
+          .sortBy("createdAt:desc")
+          .get("firstObject");
+      }
+    );
+  },
+
+  getAllOnlineOrAppointmentOrder({ appointment }) {
```
patrixr: I think the name is a bit confusing, maybe something simply having `fetchOnlineOrders()` is enough

patrixr: or `fetchOrders({ type: 'appointment' })`

patrixr: `OR` kinda implies that it will return both, which makes it confusing

app/controllers/account_details.js
```diff
@@ -85,11 +86,7 @@ export default Ember.Controller.extend({
         }
       });
     } else if (onlineOrder) {
-      this.transitionToRoute("request_purpose", {
-        queryParams: {
-          bookAppointment: false
-        }
-      });
+      this.transitionToRoute("submitted_orders");
```
patrixr: Just a suggestion, maybe renaming the page to something clearer, like `submitted_order_selection`

`submitted_orders` could imply that it's just a list (similar to the dashboard), but it's actually more a form

app/routes/orders/goods.js
```diff
@@ -23,5 +33,12 @@ export default detail.extend({
       data.packages = this.loadPackagesOf(data.order);
       return Ember.RSVP.hash(data);
     });
+  },
+
+  setupController(controller, model) {
+    this._super(controller, model);
+    if (this.get("showMessage")) {
```
patrixr: what if `controller. showUpdateMessage` is already true because of a previous order. Maybe it's worth setting it to false : )

`controller.set("showUpdateMessage", this.get("showMessage"));` would work

(although the comment I wrote above kinda makes this comment pointless :D )

app/services/order-service.js
```diff
@@ -118,12 +118,34 @@ export default ApiService.extend({
...
-        .get("firstObject");
+      return orders.filterBy("isAppointment", appointment);
     });
+  },
+
+  hasSubmittedOrders() {
+    return this.getAllOnlineOrAppointmentOrder({ appointment: false }).then(
+      orders => {
+        const hasSubmittedOrder =
+          orders.filter(this.isOrderSubmittedOrProcessed).length > 0;
```
patrixr: you can use `orders.some( ... )`

app/routes/orders/goods.js
abulsayyad123: Removed this logic and replaced it with packages recently updated time.

app/routes/submitted_orders.js
abulsayyad123: Done.

app/controllers/account_details.js
abulsayyad123: Yeah... changed its name to `sumitted_order_selection`.

PR #439, opened 2019-07-22
Handle backlink and hide prev next buttons for search page items

abhinav098:
Hello team 
This PR fixes the issue of back link on browse item page, if we click prev or next button, the back link malfunctioned. This PR will fix those issues.

Also, I've hidden `prev` and `next` button from the items page, if the user is coming from search page as I feel they don't make sense there. Let me know if I should change it back.

Thanks


app/controllers/item.js
```diff
@@ -46,8 +47,10 @@ export default Ember.Controller.extend({
   linkDisplayName: Ember.computed("prevPath", function() {
     let prevPath = this.get("prevPath");
     if (prevPath === "search_goods") {
+      this.set("showPrevNextButtons", false);
```
patrixr: we can't `set` anything in a computed getter. It is likely to cause unpredictable side-effects in the future. And we cannot guarantee that the computed property will ever be computed at the right time (if ever).

patrixr: maybe this instead: 

```javascript
showPrevNextButtons: Ember.computed("prevPath", function() {
   return prevPath !== "search_goods"
});
```

abhinav098: okay, looks better. 
Thanks :)

app/controllers/item.js
```diff
@@ -138,18 +152,15 @@ export default Ember.Controller.extend({
...
-      );
-      if (parentId) {
-        this.get("packageCategory").set("selectedCategoryId", category);
-      }
+      this.setAndRedirectToCategory(category);
     },
 
     back() {
-      window.history.back();
+      let prevPath = this.get("prevPath");
```
patrixr: would the opposite logic be possible ?

```
if (hasParentCategory) {
  return this.setAndRedirectToCategory(this.get("categoryObj"));
}
window.history.back();
```

allowing us to keep the `window.history.back()` , which we've been trying to implement.

Also if you plan on controlling the navigation yourself, you might want to enter the page using a `replace` instead of a `transition`. To avoid a large looping history

now the navigation history would look like `search_goods -> item -> search_goods`

abhinav098: @patrixr I added `window.history.back()` some time back, but according to the requirement `window.history` doesn't suits here. Because when we click on `previous` and `next` buttons on item's page the `window.history` maintains its history and backlink malfunctions.
Case 1:
What I've done here is, if a child category is chosen, back action will take back to the parent category url with child category already chosen.
Also if on the previous

abhinav098: And regarding the history, currently I have handled the history from both the pages.
If the user is on item page user will be redirected according to the logic I just explained in my above mentioned comment.

If user is on search page and he clicks on back or done, he'll be redirected to the browse-page
https://github.com/crossroads/browse.goodcity/pull/439/files#diff-44c533ec5932f27a50b79b2469b38784R77

abhinav098: > `replace` instead of a `transition`

Actually i'm doing this transition in the controller, and replaceWith is not available in the controller, for that I'll have to access routing inside of a controller, but I don't think that's a good practice. So I avoided it @patrixr

app/controllers/item.js
```diff
@@ -78,6 +82,17 @@ export default Ember.Controller.extend({
     }
   }),
 
+  setAndRedirectToCategory(category) {
+    const parentId = category.get("parentId");
+    this.transitionToRoute(
+      "package_category",
+      parentId ? parentId : category.id
+    );
+    if (parentId) {
```
abulsayyad123: I guess this should have been written before transition, any reason to write it below ?

app/templates/item.hbs
```diff
@@ -16,21 +16,23 @@
...
-            <a disabled="{{noNextItem}}" class="no-avatar">
-              <span class="hide-for-small {{if hasCartItems 'narrow' ''}}">{{t "itemdetail.next"}}</span>
-              <i class="fa fa-arrow-right"></i>
-            </a>
-          </li>
-        </ul>
+        {{#if showPrevNextButtons}}
+          <ul class="text-right pagination right">
+            <li class="arrow prev enabled-arrow" {{action "setDirectionAndRender" "moveRight" on="click"}}>
+              <a disabled="{{noPreviousItem}}" class="no-avatar">
```
abulsayyad123: did this passing binding within quotes really worked  i.e `disabled = "{{noPrevousItem}}"` ?

abhinav098: yes it works. Existing code.

app/controllers/item.js
abhinav098: no reason, works bothways.

PR #443, opened 2019-07-24
Fixes : cart widget + cart sidebar

patrixr:
Changes:

- Add `Show order summary` widget back into the order process
- Fix the cart item list within the widget
- Fix issue where clicking on `show` on the car sidebar right after requesting an item would close the sidebar instead of expanding it.

### DIFF
--- a/app/components/small-cart-toggle-view.js
+++ b/app/components/small-cart-toggle-view.js
@@ -2,10 +2,21 @@ import Ember from "ember";
 
 export default Ember.Component.extend({
   showCartSummary: false,
+  allowClick: false,
+
+  application: Ember.computed(function() {
+    return Ember.getOwner(this).lookup("controller:application");
+  }),
 
   actions: {
     toggleCartSummary() {
       this.toggleProperty("showCartSummary");
+    },
+    showItemDetails(item) {
+      if (!this.get("allowClick")) {
+        return;
+      }
+      this.get("application").send("showItemDetails", item);
     }
   }
 });
--- a/app/components/small-cart-toggle-view.js
+++ b/app/components/small-cart-toggle-view.js
@@ -1,12 +1,12 @@
 import Ember from "ember";
 
 export default Ember.Component.extend({
-  showCartSummary: false,
-  allowClick: false,
+  cart: Ember.inject.service(),
 
-  application: Ember.computed(function() {
-    return Ember.getOwner(this).lookup("controller:application");
-  }),
+  // Set to true to enable navigation to the details
+  // Usage: {{small-cart-toggle-view allowClick=true}}
+  allowClick: false,
+  showCartSummary: false,
 
   actions: {
     toggleCartSummary() {
@@ -16,7 +16,7 @@ export default Ember.Component.extend({
       if (!this.get("allowClick")) {
         return;
       }
-      this.get("application").send("showItemDetails", item);
+      this.get("cart").navigateToItemDetails(item);
     }
   }
 });
--- a/app/controllers/application.js
+++ b/app/controllers/application.js
@@ -173,18 +173,7 @@ export default Ember.Controller.extend(cancelOrderMixin, {
     },
 
     showItemDetails(record) {
-      let isItem = record.get("isItem");
-      let categoryId = record.get("allPackageCategories.firstObject.id");
-      let sortBy = "createdAt:desc";
-
-      const route = isItem ? "item" : "package";
-      const routeId = record.get("id");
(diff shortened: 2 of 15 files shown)


app/components/small-cart-toggle-view.js
```diff
@@ -2,10 +2,21 @@ import Ember from "ember";
 
 export default Ember.Component.extend({
   showCartSummary: false,
+  allowClick: false,
+
+  application: Ember.computed(function() {
+    return Ember.getOwner(this).lookup("controller:application");
```
Darshan-Chauhan: I don't think that it's a good idea to bind component with controller. 
You can inject a controller in a controller but ember specifically doesn't allow you to inject a controller in a component. 

Is there any other or better way to do this?

patrixr: It's definitely not, but I'm trying to find the path of least resistance.
We have 4 or 5 different implementations of the cart, with just enough differences to make it hard to use a single component everywhere. The action is in the application, so my thought was either re-write it here, or call the one from the application.

I think both ways are equally bad here :/

patrixr: any ideas ?

app/routes/application.js
```diff
@@ -102,7 +102,7 @@ export default Ember.Route.extend({
     if (error && _.isString(error)) {
       return error;
     }
-    return reason.errors[0].status == 403
+    return _.get(reason, "errors[0].status") == 403
```
abhinav098: @patrixr why have we changed this ? I think the `reason` is passed to this function right ?

app/components/small-cart-toggle-view.js
patrixr: The main point is that I don't want to have yet another implementation of the list, which is why I'm using  `{{partial "cart_content/items"}}` inside the component

app/routes/application.js
patrixr: This was breaking when reason didn't have an `error` property.
Using `_.get` does the same thing excepts it won't crash if a nested property doesn't exist

patrixr: It's the equivalent of doing `reason && reason.errors && reason.errors[0] && reasons.errors[0].status`

app/controllers/application.js
```diff
@@ -40,7 +40,6 @@ export default Ember.Controller.extend(cancelOrderMixin, {
     return ["browse", "package_category"].indexOf(this.get("currentPath")) >= 0;
   }),
 
-  displayCart: false,
```
abhinav098: controllers/item.js has this action
```
    requestItem(item) {
      this.get("cart").add(item);
      Ember.run.later(
        this,
        function() {
          this.get("application").send("displayCart");
        },
        50
      );
    },
```

I hope removing this wont affect this action @patrixr

app/components/small-cart-toggle-view.js
```diff
@@ -2,10 +2,21 @@ import Ember from "ember";
...
+  }),
 
   actions: {
     toggleCartSummary() {
       this.toggleProperty("showCartSummary");
+    },
+    showItemDetails(item) {
+      if (!this.get("allowClick")) {
```
Darshan-Chauhan: Ammm what is the use of `allowClick` here?
The initial value if defaulting to false and I don't see it changing anywhere else in this component.
So `!this.get("allowClick")` will be always true

Darshan-Chauhan: Also that property is not passed to component either *{{{{small-cart-toggle-view}}}}*

patrixr: It's for the ability to turn it on later on. We had a long discussion with Matt on whether we wanted to allow user to navigate away from the order process in the middle (by clicking on an item)

patrixr: We can remove it if you don't like having an un-used feature, I don't mind

app/components/small-cart-toggle-view.js
Darshan-Chauhan: Will it be possible to move ** showItemDetails** action from application controller to `cart-service` and inject that service in application controller(might be there already) and component?

app/components/small-cart-toggle-view.js
Darshan-Chauhan: I wasn't aware about the discussion with Matt 😄 
I don't mind keeping un-used features but it just hard to understand at first glance.
Can you add a similar comment so it'll be easier to understand for everyone?

patrixr:
@Darshan-Chauhan Updated 👍

PR #450, opened 2019-08-02
GCW-GCW-2728- Order state change do not update order selection page.

abulsayyad123:
Hi Team,
This PR is fix for issue on Order Selection page. When user changes Order state from "Processing" to "Awaiting Dispatch" in Stock App. This does not update Order Selection page where cart items are merged with existing Orders.
Please review the PR and let me know if any change is required.
Thanks.

### DIFF
--- a/app/controllers/submitted_order_selection.js
+++ b/app/controllers/submitted_order_selection.js
@@ -15,6 +15,32 @@ export default Ember.Controller.extend(asyncTasksMixin, {
     return this.get("cart.isEmpty");
   },
 
+  submittableOrders: Ember.computed("model.@each.state", function() {
+    let submittableOrders = this.get("model").filter(
+      this.isPackageSubmittableOrder
+    );
+    if (submittableOrders.length) {
+      return submittableOrders;
+    } else {
+      this.redirectToRequestPurpose();
+    }
+  }),
+
+  redirectToRequestPurpose() {
+    this.transitionToRoute("request_purpose", {
+      queryParams: {
+        onlineOrder: true,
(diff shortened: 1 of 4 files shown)


app/controllers/submitted_order_selection.js
```diff
@@ -15,6 +15,32 @@ export default Ember.Controller.extend(asyncTasksMixin, {
...
   },
 
+  submittableOrders: Ember.computed("model.@each.state", function() {
+    let submittableOrders = this.get("model").filter(
+      this.isPackageSubmittableOrder
+    );
+    if (submittableOrders.length) {
+      return submittableOrders;
+    } else {
+      this.redirectToRequestPurpose();
```
patrixr: redirection inside a computed property is a time bomb.  You can move that to the route, `submittableOrders` should behave like a readonly variable, thus having no side effects

app/controllers/submitted_order_selection.js
```diff
@@ -15,6 +15,32 @@ export default Ember.Controller.extend(asyncTasksMixin, {
...
+    }
+  }),
+
+  redirectToRequestPurpose() {
+    this.transitionToRoute("request_purpose", {
+      queryParams: {
+        onlineOrder: true,
```
patrixr: one of these days we'll need to clean this a bit, having 2 opposite flags `onlineOrder/bookAppointment` is bound to eventually cause us problems (maybe we can move to a `orderType: "appointment"` format).

This is unrelated to this PR, just thought I'd mention it 😄

PR #462, opened 2019-09-17
fix rollbar #612

abhinav098:
Hello team, 
This PR fixes a rollbar issue which was observed on 17th Sept,2019.
Rollbar link: https://rollbar.com/crossroads_it/browse.goodcity/items/612/

Please review 
Thanks.

### DIFF
--- a/app/services/logger.js
+++ b/app/services/logger.js
@@ -5,6 +5,11 @@ export default Ember.Service.extend({
   session: Ember.inject.service(),
   rollbar: Ember.inject.service(),
 
+  raiseError: function(errorData, currentUser) {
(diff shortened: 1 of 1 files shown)


app/services/logger.js
```diff
@@ -5,25 +5,33 @@ export default Ember.Service.extend({
   session: Ember.inject.service(),
   rollbar: Ember.inject.service(),
 
+  raiseError: function(errorData, reason, currentUser) {
```
Darshan-Chauhan: You can directly use `raiseError(arguments)`.. No need use function keyword

abhinav098: yeah sure @Darshan-Chauhan.

PR #470, opened 2019-10-07
Fixing deprecations

abhinav098:
### What does this PR do?

This PR fixes two deprecations

1) ```ember-cli-babel 5.x has been deprecated. Please upgrade to at least ember-cli-babel 6.6. Version 5.2.8 located: browse -> ember-cli-content-security-policy -> ember-cli-babel```

2) `baseURL` is deprecated. Use `rootURL` instead

### DIFF
--- a/config/environment.js
+++ b/config/environment.js
@@ -5,7 +5,7 @@ module.exports = function(environment) {
   var ENV = {
     modulePrefix: "browse",
     environment: environment,
-    baseURL: "/",
+    rootURL: "/",
     defaultLocationType: "auto",
 
     emberRollbarClient: {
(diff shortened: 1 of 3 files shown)


Darshan-Chauhan:
I would suggest doing one round of testing for Cordova iOS and Android app for this change,

abhinav098:
sure @Darshan-Chauhan

Darshan-Chauhan:
Also, what would it take to upgrade ember-cli-babel to the latest version so that we can use latest features of ES10

abhinav098:
I need to check that. Current available version is `7.2.0` and our app's version is `6.6.0`. We need to check where it might impact the browse app.

namrataukirde:
@abhinav098 Any update regarding this PR?

PR #495, opened 2020-01-14
hide order summary in case of book appointment

akshaygwd:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-2953

### What does this PR do?

Bugs:  This PR hides the order summy if the user clicks on the book appointment button to get a package.

Please review:

### DIFF
--- a/app/controllers/order/client_information.js
+++ b/app/controllers/order/client_information.js
@@ -38,6 +38,10 @@ export default Controller.extend(cancelOrder, {
     return config.APP.HK_COUNTRY_CODE + this.get("mobilePhone");
   }),
 
+  isAppointment: computed("order", function() {
+    return this.get("order.isAppointment");
+  }),
+
   titles: computed(function() {
     let translation = this.get("i18n");
     let mr = translation.t("account.user_title.mr");
--- a/app/controllers/order/client_information.js
+++ b/app/controllers/order/client_information.js
@@ -38,9 +38,7 @@ export default Controller.extend(cancelOrder, {
     return config.APP.HK_COUNTRY_CODE + this.get("mobilePhone");
   }),
 
-  isAppointment: computed("order", function() {
-    return this.get("order.isAppointment");
-  }),
+  isAppointment: alias("order.isAppointment"),
 
   titles: computed(function() {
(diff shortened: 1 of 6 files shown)


app/controllers/order/client_information.js
```diff
@@ -38,6 +38,10 @@ export default Controller.extend(cancelOrder, {
     return config.APP.HK_COUNTRY_CODE + this.get("mobilePhone");
   }),
 
+  isAppointment: computed("order", function() {
```
abulsayyad123: Use alias here

app/templates/request_purpose.hbs
```diff
@@ -34,7 +34,9 @@
             <section
               class="main-section request-purpose medium-10 medium-offset-1 large-10 large-offset-1 small-10 small-offset-1 columns">
               {{#unless cart.isEmpty}}
-              {{small-cart-toggle-view}}
+               {{#unless bookAppointment}}
```
abhinav098: @akshaygwd use the `order_summary` partial here as well.

abhinav098: `bookAppointment` where's this property defined @akshaygwd ?

akshaygwd: isAppointment property is not accessible in request_purpose.hbs file that why using bookAppointment query param for request_purpose.hbs file

abhinav098: still you can use the partial.

abhinav098: @akshaygwd use this in the partial `{{#unless is-or(isAppointment bookAppointment) }}`

so it will check for both properties. 

and you can use this partial in request-purpose as well.

cc- @abulsayyad123 @namrataukirde

akshaygwd: okay got it

abhinav098: Never mind, now that the PR is merged, lets keep it as is and avoid the tight coupling that above suggestion might cause in the partial file.

akshaygwd: sure @abhinav098

PR #499, opened 2020-01-22
Gcw 2980 account details page not visible ie

akshaygwd:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-2980

### What does this PR do?

Bug:
This PR fixes the issue of 
1. when new user login he is not redirected to the accout_detail page on ie.
2. On IE11, on the account details page for a new user, the email field contains '[object object]'

could you please review.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -355,28 +355,28 @@ workflows:
               # - test_cases
             filters:
               branches:
-                only: /^(master|live)$/
+                only: /^(master|live|GCW-2980-account-details-page-not-visible-ie)$/
         - www-deploy:
             requires:
               - www_build
             filters:
               branches:
-                only: /^(master|live)$/
+                only: /^(master|live|GCW-2980-account-details-page-not-visible-ie)$/
         - ember_cordova_build:
             requires:
               - package_dependencies
             filters:
               branches:
-                only: /^(master|live)$/
+                only: /^(master|live|GCW-2980-account-details-page-not-visible-ie)$/
         - android_build_and_deploy:
             requires:
               - ember_cordova_build
             filters:
               branches:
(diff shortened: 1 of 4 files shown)


app/routes/post_login.js
```diff
@@ -48,7 +44,7 @@ export default Route.extend({
         this.transitionTo("browse");
       }
     } else {
-      this.transitionTo("account_details");
+      window.location.replace("/account_details");
```
abhinav098: @akshaygwd, please test it on all platforms. Transition is required. I would recommend to use `transitionTo` here.

app/routes/post_login.js
```diff
@@ -15,10 +15,6 @@ export default Route.extend({
     this.set("isBookAppointment", params.queryParams.bookAppointment);
   },
 
-  model() {
-    return this.get("preloadService").preloadData();
-  },
```
abhinav098: Why has this been removed ? `model` is required. right ? Have you tested this thoroughly on local before creating the PR ?

akshaygwd: model is not used for post-login that's why I removed it

.circleci/config.yml
```diff
@@ -355,28 +355,28 @@ workflows:
...
               branches:
-                only: /^(master|live)$/
+                only: /^(master|live|GCW-2980-account-details-page-not-visible-ie)$/
         - www-deploy:
             requires:
               - www_build
             filters:
               branches:
-                only: /^(master|live)$/
+                only: /^(master|live|GCW-2980-account-details-page-not-visible-ie)$/
```
namrataukirde: Make sure you will remove `GCW-2980-account-details-page-not-visible-ie` before merging PR.

PR #501, opened 2020-01-23
Gcw 2984 dimensions are shown as null

akshaygwd:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-2984

### What does this PR do?

BUG:
This PR hides the size field from the front end if anyone of the following length X width X height is not present as all three fields are required to fill.

before:
Screen Shot 2020-01-23 at 3 44 43 PM

Please review:

### DIFF
--- a/app/models/package.js
+++ b/app/models/package.js
@@ -1,4 +1,4 @@
-import { bool, alias } from "@ember/object/computed";
+import { bool, alias, and } from "@ember/object/computed";
 import { once } from "@ember/runloop";
 import { observer, computed } from "@ember/object";
 import { getOwner } from "@ember/application";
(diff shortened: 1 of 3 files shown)


app/models/package.js
```diff
@@ -104,6 +104,10 @@ export default Model.extend(cloudinaryImage, {
     return !res ? "" : res + " cm";
   }),
 
+  isDimensionPresent: computed("width", "height", "length", function() {
```
abulsayyad123:
```suggestion
  isDimensionPresent: and("width", "height", "length")
```
You can try this one.

akshaygwd: Okay got it

PR #504, opened 2020-01-30
Jan2020 Release#1

akshaygwd:
This PR contains:

1. GCW-2888 the links on the FAQs page don't work on mobiles.
2. GCW-2901 ember upgrade to 2.18.
3. GCW-2062 Browse design clearer dimensions UI.
4. GCW-2912 A few issues on the 'Goods Details' page while setting up an appointment.
5. GCW-2896 Issues when scrolling on a short browser.
6. Fixed Rollbar environment.
7. upgrade the Fastlane version.
8. GCW-2953 show order summary should not be visible for appointment order.
9. GCW-2954 on iPhone Done button is hard to click.
10. GCW-2984 dimensions are shown as null in UI if the item does not have dimensions.
11. GCW-2994 Find some items to request not working.

Please review.


app/templates/order/_order_summary.hbs
```diff
@@ -0,0 +1,5 @@
+{{#unless cart.isEmpty}}
+  {{#unless isAppointment}}
+    {{small-cart-toggle-view}}
+  {{/unless}}
+{{/unless}}
```
abhinav098: @akshaygwd , you can use `is-or` helper that we have to avoid nested conditionals here.

akshaygwd: @abhinav098 I want both conditions to be satisfied.

abhinav098: in that case we have `is-and`.

abhinav098: @akshaygwd in that case we have `is-and`.

namrataukirde: Hi, @abhinav098 Thanks for your suggestions. Appreciate it. 

You can ask @akshaygwd to correct this in another PR as it is not completely related to this PR and can be done separately as an improvement to our code.

It's really good to mention such things so that we can improve our codebase constantly. :)

@akshatgwd please keep a note of this suggestion and fix it whenever you get time.

Thanks.

akshaygwd: @namrataukirde yeah sure.

PR #507, opened 2020-02-06
Fix broken email field

patrixr:
Moving this urgently to prod !

The email field was broken due to `value=email` appearing twice in the ember component


patrixr:
Note: merging straight away in order to create the release PR

PR #509, opened 2020-02-19
Gcw 2985 Fixed addition of zero quantity item in cart

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-2985

### What does this PR do?
This PR fixes two issue:

1. In the Browse app, it does not fetch the items with zero quantity in the response. But there are some cases where item page is open in browser and it is dispatched from the stock app, which changes the quantity to zero but the `Request Item` button is still active.
Fixed this issue by disabling the `Request Item` button 

Screen Shot 2020-02-18 at 12 16 04 PM

2. On cart show page `/cart`, even though there are no items in cart, `Submit order` button is active and which shows message on click as 
Screen Shot 2020-02-18 at 12 21 21 PM

This seems bit confusing, as User have not added any item at all. 

Fixed it by disabling the `submit order` button.

Screen Shot 2020-02-19 at 9 09 10 AM

### Impacted Areas

`/cart` and item page

### DIFF
--- a/app/controllers/item.js
+++ b/app/controllers/item.js
@@ -42,8 +42,10 @@ export default Controller.extend({
       : [item];
   }),
 
-  notAvailableInStock: computed("allPackages.@each.quantity", function() {
-    let quantities = this.get("allPackages").map(pkg => pkg.get("quantity"));
+  notAvailableInStock: computed("allPackages.@each.onHandQuantity", function() {
+    let quantities = this.get("allPackages").map(pkg =>
+      pkg.get("onHandQuantity")
+    );
     return _.sum(quantities) === 0;
   }),
 
--- a/app/controllers/item.js
+++ b/app/controllers/item.js
@@ -4,6 +4,7 @@ import { alias, empty, gt, sort } from "@ember/object/computed";
 import { inject as service } from "@ember/service";
 import Controller, { inject as controller } from "@ember/controller";
 import config from "../config/environment";
+import _ from "lodash";
 
 export default Controller.extend({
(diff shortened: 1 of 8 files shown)


app/controllers/item.js
```diff
@@ -41,6 +42,11 @@ export default Controller.extend({
       : [item];
   }),
 
+  notAvailableInStock: computed("allPackages.@each.quantity", function() {
+    let quantities = this.get("allPackages").map(pkg => pkg.get("quantity"));
+    return _.sum(quantities) === 0;
+  }),
+
```
patrixr: We're in the process of removing the `quantity` field, as it is faulty. So if you could avoid using it that'd be very helpful :)
As of now the available quantity can still be inferred from the `packages_locations` and the `orders_packages`.

In the near future we will have multiple fields like `available_quantity`, `designated_quantity`, etc... (IN PROGRESS)

abhinav098: @swatijadhav , you can check the `stockit_item_serializer`(`on_hand_attribute`) to check how we're calculating the available quantity. 

Unfortunately, we are not using `stockit_item_serializer` on browse. So you can check how we can send the `on_hand_quantity` to browse using the `package_serializer`. 

Using quantity from now on can become misleading in the coming future and we might need to fix that throughout. As @patrixr mentioned , we want to get rid of this quantity column.

patrixr: If the records are available on the frontend then a computed property could be enough as a temporary measure, rather than making more api changes

PR #511, opened 2020-02-20
Release send booking type for appointment slots

patrixr:
This is related to an urgent request from the staff to lock down only appointments during the Coronavirus outbreak.

It will give the API more control over the dates that it sends


patrixr:
> just a reminder, to change version on CircleCI as well.

Already done ;)

PR #512, opened 2020-02-20
 GCW-2920 avoid creation of empty order

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-2920

### What does this PR do?

**Issue 1:**
When on click of `Request Item` gives any error for any edge case, it still adds that item to cart. And allows to create order and eventually creates order with no item.

Fix: Unload requested-package record from cart if receive any error from backend

**Issue 2**
For logged-in users, while cart submit, it does not check item's availability. And allows to create order with no item.

Fix: For logged-in user, updated condition to check for cart having items and availability of those items. 

### Impacted Areas

Cart page

### DIFF
--- a/app/controllers/application.js
+++ b/app/controllers/application.js
@@ -49,6 +49,7 @@ export default Controller.extend(cancelOrderMixin, {
   cartscroll: service(),
 
   hasCartItems: alias("cart.isNotEmpty"),
+
   cartLength: alias("cart.groupedPackages.length"),
 
   isUserLoggedIn: computed("session.authToken", function() {
@@ -169,7 +170,7 @@ export default Controller.extend(cancelOrderMixin, {
       const accountComplete = this.get("session").accountDetailsComplete();
       const loggedIn = this.get("session.isLoggedIn");
 
-      if (loggedIn && !accountComplete && this.get("hasCartItems")) {
+      if (loggedIn && !accountComplete && this.get("cart.hasValidItems")) {
         this.set("showCartDetailSidebar", false);
         return this.transitionToRoute("account_details", {
           queryParams: {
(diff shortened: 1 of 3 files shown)


app/services/cart.js
```diff
@@ -180,6 +181,7 @@ export default ApiService.extend({
    * If the user is not logged in, will only save it locally
    */
   addPackage(pkg) {
+    let _this = this;
```
abulsayyad123: No need of creating `_this`. It will be handled since `arrow` function is used in `catch` block.

app/controllers/application.js
```diff
@@ -49,6 +49,18 @@ export default Controller.extend(cancelOrderMixin, {
   cartscroll: service(),
 
   hasCartItems: alias("cart.isNotEmpty"),
+
+  hasValidCartItems: computed(
```
abulsayyad123: was thinking can we have this computed property in `cart` service itself. This way we can just use
 `cartService. hasValidCartItems`. Thoughts ?

app/controllers/application.js
```diff
@@ -49,6 +49,18 @@ export default Controller.extend(cancelOrderMixin, {
...
+    "cart.unavailableCartItems",
+    "cart.isNotEmpty",
+    function() {
+      return (
+        this.get("cart.isNotEmpty") &&
+        this.get("unavailableCartItems.length") === 0
```
namrataukirde: I think we don't need to add equality check here,  just use of length is enough. 0 is considered false.

`var a = ""
a.length ? "a" : "b"
=> "b"
var a = "hello"
a.length ? "a" : "b"
=> "a"`

swatijadhav: Since using this value for `disabled` attribute in the handlebar template, when evaluated to `0` it sets the value as `disabled="0"` and it just disables the button. Hence passed explicit check which will either return a true or false value.

namrataukirde: Okay. Sounds Good :+1: :)

PR #522, opened 2020-03-03
GCW-3070 fix for page breaking if package is not present

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3070

### What does this PR do?
This PR will fix bug in browser app when a particular item is viewed when it is not available in the inventory.

Steps to reproduce:
1. View a package in the browse app
2. Designate all the quantities for the same package
3. Refresh the page in the browse app, the page breaks

### Impacted Areas

Browse > Category

### Screenshots

Screen Shot 2020-03-03 at 9 53 17 AM

### DIFF
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -142,6 +142,7 @@ export default Route.extend({
 
   handleError: function(reason) {
     try {
+      debugger;
       var status;
       // let hasPopup = Ember.$('.reveal-modal:visible').length > 0;
       try {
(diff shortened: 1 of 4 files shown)


app/routes/application.js
```diff
@@ -180,7 +180,11 @@ export default Route.extend({
       // as we're trying to render a page and remove loading
       // indicator at a same time
       later(() => {
-        if (this.loadingView) {
+        // Clear the loader only if the loader is actually present in the DOM
+        if (
+          this.loadingView &&
+          document.getElementById(this.loadingView.elementId)
```
abulsayyad123: Do we really need DOM check here ?  I think check on `this.loadingView.elementId` is enough. Thoughts ?

PR #530, opened 2020-03-13
GCW-2997 add loading screen at request and remove item

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-2997

### What does this PR do?
Add loading screen at the request and remove the item.

### DIFF
--- a/app/services/cart.js
+++ b/app/services/cart.js
@@ -4,7 +4,7 @@ import { computed } from "@ember/object";
 import { inject as service } from "@ember/service";
 import ApiService from "./api-base-service";
 import _ from "lodash";
-import { getOwner } from "@ember/application";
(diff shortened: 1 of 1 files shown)


app/services/cart.js
```diff
@@ -216,10 +227,20 @@ export default ApiService.extend({
   },
 
   remove(pkgOrItem) {
+    let promise;
+    let loadingView = getOwner(this)
```
abulsayyad123: We can use `runTask` from `asyn_task` mixins.

abulsayyad123: Eg: https://github.com/crossroads/browse.goodcity/blob/master/app/controllers/order/confirm_booking.js#L16

PR #543, opened 2020-04-02
GCW-2990 fix the something went wrong error on order cancellation

bharat619:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-2990

### What does this PR do?
Resolves the error for `Something went wrong`

### DIFF
--- a/app/controllers/order/client_information.js
+++ b/app/controllers/order/client_information.js
@@ -1,5 +1,6 @@
 import $ from "jquery";
 import { computed } from "@ember/object";
+import { inject as service } from "@ember/service";
 import { alias, equal } from "@ember/object/computed";
 import Controller, { inject as controller } from "@ember/controller";
 import { getOwner } from "@ember/application";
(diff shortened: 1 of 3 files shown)


app/controllers/order/client_information.js
```diff
@@ -90,52 +92,58 @@ export default Controller.extend(cancelOrder, {
...
-    } else {
-      store.pushPayload(beneficiaryResponse);
+      if (beneficiary && actionType === "DELETE") {
+        store.unloadRecord(beneficiary);
+      } else {
+        store.pushPayload(beneficiaryResponse);
+      }
+      this.send("redirectToGoodsDetails");
+      loadingView.destroy();
+    } catch (err) {
```
abulsayyad123, line 142: I think it would be better to use send `runTask` mixins. It will automatically handle error instead of explicitly using `messageBox` here.

steveyken:
@bharat619 I know this ticket has been parked for time being... do you want to convert this PR to draft? It will help clear out the PR review queue until it's ready

PR #549, opened 2020-04-09
GCW-3084 removed old previous-route related code changes and added new back-link component

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-3084

### What does this PR do?
Removed old code of tracking back route which is not working after upgrade of ember as those features are removed.
Added changes for back link to previous pages.

### DIFF
--- a/app/components/back-link.js
+++ b/app/components/back-link.js
@@ -1,0 +1,27 @@
+import Component from "@ember/component";
+import config from "browse/config/environment";
+
+export default Component.extend({
+  fallbackRoute: "",
+
+  actions: {
+    back() {
+      const fallbackRoute = this.get("fallbackRoute") || "/";
+      const isRoot = location.pathname === "/";
+
+      if (isRoot) {
+        return;
+      }
+
+      const noHistory = [1, 2].includes(window.history.length);
+      const deepLinked =
+        document.referrer && document.referrer.indexOf(config.APP.ORIGIN) < 0;
+
+      if (deepLinked || noHistory) {
+        return this.get("router").replaceWith(fallbackRoute);
+      }
+
+      window.history.back();
+    }
+  }
+});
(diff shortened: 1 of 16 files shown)


app/routes/order/goods_details.js
```diff
@@ -10,23 +10,6 @@ export default AuthorizeRoute.extend({
     fromClientInformation: false
   },
 
-  beforeModel() {
-    var previousRoutes =
```
patrixr, line 14: so happy seeing this gone :D

app/components/back-link.js
```diff
@@ -0,0 +1,27 @@
+import Component from "@ember/component";
```
patrixr, line 1: @swatijadhav Have you encountered this flow ?

* There is a link to a goodcity page, you click on 'open in a new tab' (history.length == 1)
* You navigate to another page (history.length == 2)
* You navigate back, `noHistory` is true, so it goes back to home instead of the previous page

I don't think this should hold this ticket, I just haven't found a good way to deal with this scenario AND keep it working when we open a new tab (and history.length starts at 2). Curious to see if you have any ideas :)

swatijadhav: @patrixr 
Yes, I found that `history.length` value differs
- when opening in a new tab, history.length is 1
- when pasting URL in a new tab, history.length is 2
so in both cases, on click of the back, it will be redirected to the `fallback` route, which is fine.

Only issue is, when a user is using tab for different URLs and then paste browse app URL, at this point if he clicks on `back` link, history.length is > 2.. and here it breaks the flow and goes to `window.history.back` which is other app URL.

The solution to this is checking `document.referrer` when this value is an empty string `document.referrer === ""` that means URL is copy-pasted.
But in the recent version of ember, `link-to` is not keeping the value of `document.referrer` and hence it will be always empty value.
(This solution is used in stock app https://github.com/crossroads/stock.goodcity/pull/698/files and it works fine.)

PR #577, opened 2020-07-13
GCW-3186 Add quantity indicator for items

shreyas098:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3186

### What does this PR do?

FEATURE: Adds quantity widget feature

### Screenshots

Screen Shot 2020-07-14 at 9 57 46 AM

Screen Shot 2020-07-14 at 9 58 05 AM

### DIFF
--- a/app/components/quantity-widget.js
+++ b/app/components/quantity-widget.js
@@ -16,6 +16,11 @@ export default Component.extend({
     }
   },
 
+  validValueCheck(pkg) {
+    let quantity = +this.get("value");
+    return quantity < 1 || quantity > pkg.get("availableQuantity") || !quantity;
+  },
+
   setUpdatedValue(value, isValueValid) {
     if (isValueValid) {
       this.set("showErrorMessage", false);
@@ -29,7 +34,7 @@ export default Component.extend({
   actions: {
     incrementQty(quantity) {
       let incrementedValue = +this.get("value") + 1;
-      this.setUpdatedValue(incrementedValue, incrementedValue < +quantity);
+      this.setUpdatedValue(incrementedValue, incrementedValue <= +quantity);
     },
 
     decrementQty() {
@@ -38,17 +43,11 @@ export default Component.extend({
     },
 
     focusOut(pkg) {
-      let quantity = +this.get("value");
-      if (
-        quantity < 1 ||
-        quantity > pkg.get("availableQuantity") ||
-        !quantity
-      ) {
+      if (this.validValueCheck(pkg)) {
         this.set("showErrorMessage", true);
         this.set("value", 1);
-        quantity = 1;
       }
-      this.performAction(quantity);
+      this.performAction(+this.get("value"));
     },
 
     // Fix: Failing randomly(Known issue with Ember)
--- a/app/components/quantity-widget.js
+++ b/app/components/quantity-widget.js
@@ -3,42 +3,40 @@ import Component from "@ember/component";
 
 export default Component.extend({
   value: computed("model.package", function() {
-    if (this.get("package.requestedPackage")) {
-      return this.get("package.requestedPackage.quantity");
-    }
-    return 1;
+    return this.get("package.requestedPackage")
+      ? this.get("package.requestedQuantity")
+      : 1;
   }),
-  setValueIfValid(value, isValidValue, id) {
+
+  setValue(value, isValidValue) {
     if (isValidValue) {
       this.set("showErrorMessage", false);
-      Ember.$("#qty" + id).val(value);
+      this.set("value", value);
     } else {
       this.set("showErrorMessage", true);
     }
   },
-  elementValue(id) {
-    return Ember.$("#qty" + id).val();
-  },
+
   actions: {
-    incrementQty(quantity, id) {
-      let incrementedValue = +this.elementValue(id) + 1;
-      this.setValueIfValid(incrementedValue, incrementedValue <= +quantity, id);
+    incrementQty(quantity) {
+      let incrementedValue = +this.get("value") + 1;
+      this.setValue(incrementedValue, incrementedValue <= +quantity);
     },
 
-    decrementQty(id) {
-      let decrementedValue = +this.elementValue(id) - 1;
-      this.setValueIfValid(decrementedValue, decrementedValue >= 1, id);
+    decrementQty() {
+      let decrementedValue = +this.get("value") - 1;
+      this.setValue(decrementedValue, decrementedValue >= 1);
     },
 
     focusOut(pkg) {
-      const quantity = this.elementValue(pkg.id);
+      const quantity = +this.get("value");
       if (
         +quantity < 1 ||
         +quantity > +pkg.get("availableQuantity") ||
         !+quantity
       ) {
         this.set("showErrorMessage", true);
-        Ember.$("#qty" + pkg.id).val(1);
+        this.set("value", 1);
         return false;
       }
     },
--- a/app/components/quantity-widget.js
+++ b/app/components/quantity-widget.js
@@ -8,33 +8,29 @@ export default Component.extend({
       : 1;
   }),
 
-  actionValue(value) {
-    if (this.get("action") == "setRequestValue") {
-      this.sendAction("action", value);
+  performAction(value) {
+    if (this.get("type") == "request") {
+      this.requestAction(value);
     } else {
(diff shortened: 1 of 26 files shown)


app/components/quantity-widget.js
```diff
@@ -0,0 +1,50 @@
+import { computed } from "@ember/object";
+import Component from "@ember/component";
+
+export default Component.extend({
+  value: computed("model.package", function() {
+    if (this.get("package.requestedPackage")) {
```
bharat619, line 6: u can make use of ternary if here

app/components/quantity-widget.js
```diff
@@ -0,0 +1,50 @@
...
+      return this.get("package.requestedPackage.quantity");
+    }
+    return 1;
+  }),
+  setValueIfValid(value, isValidValue, id) {
```
bharat619, line 11: keep the name as setValue

app/components/quantity-widget.js
```diff
@@ -0,0 +1,50 @@
...
+    if (isValidValue) {
+      this.set("showErrorMessage", false);
+      Ember.$("#qty" + id).val(value);
```
bharat619, line 14: why `Ember.$("#qty" + id).val(value);`?

`this.set('value', value)` will this not work? 🤔

app/components/quantity-widget.js
```diff
@@ -0,0 +1,50 @@
...
+    } else {
+      this.set("showErrorMessage", true);
+    }
+  },
+  elementValue(id) {
+    return Ember.$("#qty" + id).val();
```
bharat619, line 20: same as above,

`return this.get('value')`

bharat619: u can completely get rid of `elementValue` property since you have `value` property

app/controllers/application.js
```diff
@@ -162,11 +162,11 @@ export default Controller.extend(cancelOrderMixin, {
       );
     },
 
-    checkout() {
+    async checkout() {
       if (this.get("cart.isEmpty")) {
         return;
       }
-
+      await this.get("cart").updateRequestedQuantity();
```
bharat619, line 169: not sure if want to await this,
`updateRequestedQuantity` is not returning any promise

app/controllers/order/confirm_booking.js
```diff
@@ -45,7 +45,7 @@ export default Controller.extend(cancelOrderMixin, asyncTasksMixin, {
       if (this.badCart()) {
         return this.i18nAlert("items_not_available", _.noop);
       }
-
+      await this.get("cart").updateRequestedQuantity();
```
bharat619, line 48: same as previous comment

app/controllers/orders/goods.js
```diff
@@ -38,11 +38,44 @@ export default detail.extend({
...
+    let res = [];
+    let refs = {};
+    ordersPackages.map(req => {
+      if (!req.get("package.hasSiblingPackages")) {
+        res.push(req);
+        return;
+      }
+      const packageSet = req.get("package.packageSet");
+      const packageSetId = packageSet.get("id");
+      refs[packageSetId];
```
bharat619, line 66: u might want to remove this line?

app/controllers/package_set.js
```diff
@@ -36,27 +38,30 @@ export default Controller.extend({
...
 
-      return record.get("isSet")
-        ? record.get("packages").filterBy("isAvailable")
-        : [record];
+      return record.get("isSet") ? record.get("packages") : [record];
     }
   ),
 
-  notAvailableInStock: computed(
+  notAvailableInStock: observer(
```
bharat619, line 54: can we avoid observer?

cc: @abulsayyad123

app/controllers/package_set.js
```diff
@@ -36,27 +38,30 @@ export default Controller.extend({
...
     "allPackages.@each.availableQuantity",
     function() {
-      let quantities = this.get("allPackages").map(pkg =>
-        pkg.get("availableQuantity")
+      let quantity = this.get("allPackages").any(
```
bharat619, line 57: shouldn't this be `.some()`? 🤔

app/controllers/application.js
shreyas098: Yes , updateRequestedQuantity is returning promise , hence await is added

app/controllers/order/confirm_booking.js
shreyas098: required..

app/controllers/package_set.js
shreyas098: In order to fix regression of live updating I have added observer which was not happening with computed..but please do let me know if there is any way to do it

app/controllers/package_set.js
shreyas098: sure..even that can be done..i will do it

app/components/quantity-widget.js
shreyas098: 👍

app/controllers/orders/goods.js
shreyas098: yes... thanks 👍

app/components/quantity-widget.js
shreyas098: It will but than i would no be able to call it in cart service.To avoid 2 way binding i have added in jquery way

app/components/quantity-widget.js
shreyas098: as mentioned above it used to avoid 2 way binding

app/components/quantity-widget.js
```diff
@@ -0,0 +1,54 @@
...
+      let incrementedValue = +this.get("value") + 1;
+      incrementedValue <= +quantity
+        ? this.set("value", incrementedValue) &&
+          this.performAction(incrementedValue)
+        : this.set("showErrorMessage", true);
+    },
+
+    decrementQty() {
+      let decrementedValue = +this.get("value") - 1;
+      decrementedValue >= 1
```
abulsayyad123, line 30: Use if-else instead of ternary to better clarity.

app/components/quantity-widget.js
```diff
@@ -0,0 +1,54 @@
...
+        ? this.set("value", decrementedValue) &&
+          this.performAction(decrementedValue)
+        : this.set("showErrorMessage", true);
+    },
+
+    focusOut(pkg) {
+      const quantity = +this.get("value");
+      if (
```
abulsayyad123, line 38: you can move code block of `if` into separate method and give it some meaningful name.

app/controllers/application.js
```diff
@@ -25,7 +26,7 @@ export default Controller.extend(cancelOrderMixin, {
   isHomePage: computed("currentPath", function() {
     return this.get("currentPath") === "home";
   }),
-
+  updatedQuantity: EmberObject.create({}),
```
abulsayyad123, line 29: Why specifically EmberObject, I think normal object could work ?

app/controllers/order/confirm_booking.js
```diff
@@ -20,6 +22,12 @@ export default Controller.extend(cancelOrderMixin, asyncTasksMixin, {
     );
   },
 
+  updateRequestedQuantityValue(record) {
```
abulsayyad123, line 25: I can see same code repeated at two places is it because you are using the counter at two places ? If that's the case move the common code into mixin.

app/models/package.js
```diff
@@ -89,6 +90,10 @@ export default Model.extend(cloudinaryImage, {
     return this.get("packageType.name");
   }),
 
+  orderedQuantity: computed("ordersPackage", function() {
+    return this.get("ordersPackage.quantity");
+  }),
```
swatijadhav, line 95:
```suggestion
  orderedQuantity: computed.alias("ordersPackage.quantity"),
```

app/components/quantity-widget.js
shreyas098: sure

app/controllers/application.js
shreyas098: Sure..done

app/controllers/order/confirm_booking.js
shreyas098: Sure

app/components/quantity-widget.js
shreyas098: Done

PR #580, opened 2020-07-29
Fixes for breaking of the live updates of quantity in quantity feature

shreyas098:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3186

### What does this PR do?
Fixes for breaking of the live updates of quantity in quantity feature and it also fix the issue mentioned in https://jira.crossroads.org.hk/browse/GCW-3031

### DIFF
--- a/app/controllers/package_set.js
+++ b/app/controllers/package_set.js
@@ -59,20 +59,6 @@ export default Controller.extend({
     }
   ),
 
-  notAvailableInStock: observer(
-    "allPackages.@each.availableQuantity",
-    function() {
-      let quantity = this.get("allPackages").any(
-        pkg => pkg.get("availableQuantity") == 0
-      );
-      if (quantity) {
-        this.set("packageUnavailableInSet", true);
-      } else {
-        this.set("packageUnavailableInSet", false);
-      }
-    }
-  ),
-
   categoryObj: computed("categoryId", function() {
     if (this.get("categoryId")) {
       return this.store.peekRecord("package_category", this.get("categoryId"));
--- a/app/controllers/package_set.js
+++ b/app/controllers/package_set.js
@@ -55,7 +55,6 @@ export default Controller.extend({
       let quantity = this.get("allPackages").any(
         pkg => pkg.get("availableQuantity") == 0
       );
-      console.log(quantity);
       return quantity ? true : false;
     }
   ),
--- a/app/controllers/package_set.js
+++ b/app/controllers/package_set.js
@@ -55,7 +55,7 @@ export default Controller.extend({
       let quantity = this.get("allPackages").any(
         pkg => pkg.get("availableQuantity") == 0
       );
-      return quantity ? true : false;
+      return !!quantity;
     }
   ),
 
--- a/app/controllers/package_set.js
+++ b/app/controllers/package_set.js
@@ -49,6 +49,16 @@ export default Controller.extend({
       return record.get("isSet") ? record.get("packages") : [record];
     }
   ),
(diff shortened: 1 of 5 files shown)


app/services/cart.js
```diff
@@ -151,6 +151,7 @@ export default ApiService.extend(asyncMixin, {
    */
   refresh() {
     this.get("store").unloadAll("requested_package");
+    this.get("store").findAll("package", { reload: true });
```
abulsayyad123, line 154: What is the reason to fetch `package` here ?

shreyas098: To get the remaining quantities of the packages after the order is placed

app/controllers/package_set.js
```diff
@@ -49,6 +49,15 @@ export default Controller.extend({
       return record.get("isSet") ? record.get("packages") : [record];
     }
   ),
+  packageUnavailableInSet: computed(
+    "allPackages.@each.availableQuantity",
+    function() {
+      let quantity = this.get("allPackages").any(
+        pkg => pkg.get("availableQuantity") == 0
+      );
+      return quantity ? true : false;
```
abulsayyad123, line 58: `!!quantity` would be better.

app/services/cart.js
```diff
@@ -124,6 +124,11 @@ export default ApiService.extend(asyncMixin, {
       order_id: order.get("id")
     }).then(data => {
       this.get("store").pushPayload(data);
+      data.orders_packages.map(data => {
+        this.get("store").findRecord("package", data.package_id, {
```
abulsayyad123, line 128:
```
   data.orders_packages.map(({package_id}) => this.get("store").peekRecord("package", package_id);
```
Can we try something like this ?

app/controllers/package_set.js
bharat619: its a boolean function.
You can name it something with `isXYZ` kind of name. For example: `isPackageUnavailableInSet`

Further you can make it as
```javascript
isPackageUnavailableInSet: computed(
    "allPackages.@each.availableQuantity",
    () => this.get("allPackages").any(pkg => !pkg.get("availableQuantity"))
)
```

app/controllers/package_set.js
```diff
@@ -49,6 +49,15 @@ export default Controller.extend({
...
+    "allPackages.@each.availableQuantity",
+    function() {
+      let quantity = this.get("allPackages").any(
+        pkg => pkg.get("availableQuantity") == 0
+      );
+      return !!quantity;
+    }
+  ),
 
   notAvailableInStock: observer(
```
patrixr, line 62: Since you are now computing `packageUnavailableInSet`, you should probably remove `notAvailableInStock` since it's manually setting that property in an observer.

I'm not sure why it was an observer to begin with, computed property is definitely the way to go 👍

app/services/cart.js
```diff
@@ -124,6 +124,11 @@ export default ApiService.extend(asyncMixin, {
...
+          reload: true
+        });
+      });
```
patrixr, line 131: Big red flag here, you're creating promises via `findRecord` without returning them or handling failures. That should almost never happen

```suggestion
      return data.orders_packages.map(data => {
        return this.get("store").findRecord("package", data.package_id, {
          reload: true
        });
      });
```

app/controllers/package_set.js
shreyas098: sure..done

app/services/cart.js
shreyas098: sure

PR #603, opened 2020-12-03
GCW-3430 Goods Detail

bharat619:
### What does this PR do?
- Avoids creation of empty goodcity_request
- Removed search_code route, and added overlay instead
- Refactor


shreyas098 (approved):

app/components/package-type-overlay.js
```diff
@@ -18,17 +17,19 @@ export default Controller.extend({
...
   }),
 
   hasSearchText: computed("searchText", function() {
-    return $.trim(this.get("searchText")).length;
+    return this.get("searchText").trim().length;
   }),
 
   hasFilter: computed("filter", function() {
-    return $.trim(this.get("filter")).length;
+    return this.get("filter").trim().length;
```
abulsayyad123, line 32: much better 👍

app/controllers/order/goods_details.js
```diff
@@ -1,82 +1,99 @@
...
+              } else {
+                params.id = gr.id;
+                return this.updateGoodsDetails(orderId, params);
+              }
+            })
+          )
+        );
+        this.transitionToRoute("order.schedule_details", orderId);
+      } catch (e) {
+        this.get("messageBox").alert(e.responseJSON.errors[0].message);
```
abulsayyad123, line 95: Since this is multiple promises, I guess if there are multiple errors you are catching only the first ones.

app/services/order-service.js
```diff
@@ -186,5 +186,37 @@ export default ApiService.extend({
...
+        quantity: params.quantity,
+        order_id: orderId,
+        package_type_id: params.packageTypeId
+      }
+    });
+    this.get("store").pushPayload(data);
+    return data;
+  },
+
+  async updateGoodsDetails(orderId, params) {
```
abulsayyad123, line 204: can we create a separate method for the common code.
```
_makeRequest(action, url , params) {
   //move common logic here
   return data;
}
```
In createGoodsDetails
```
async createGoodsDetails(orderId, params) {
    const url = `/goodcity_requests`;
    return _makeRequest('POST', url, params)
 }
```
Similary for Update as well

Just a suggestion.

abulsayyad123 (approved):
Just verify about the error catching, one improvement suggestion, rest looks good 👍

PR #608, opened 2020-12-23
GCW-3405-User can responds to offers

abulsayyad123:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3405

### What does this PR do?

This PR adds offers listing shared from admin API. It is using the `V2` API. It has the following pages:
- Offers list
- Detail offer page
- Chat page for the offer.

### DIFF
--- a/app/components/item-thumbnail-view.js
+++ b/app/components/item-thumbnail-view.js
@@ -1,0 +1,45 @@
+import { bind } from "@ember/runloop";
+import Component from "@ember/component";
+import { inject as service } from "@ember/service";
+import cloudinaryImage from "../mixins/cloudinary_image";
+
+export default Component.extend(cloudinaryImage, {
+  loading: true,
+  store: service(),
+
+  changedSrc: function() {
+    if (!this.is_cached(this.get("src"))) {
+      this.set("loading", true);
+    }
+  }.observes("src"),
+
+  didReceiveAttrs() {
+    this._super(...arguments);
+    this.set("cloudinaryId", this.get("cloudinaryId"));
+    this.set("src", this.generateUrl(500, 500, true));
+    const pkgTypeType = this.get("store").peekRecord(
+      "package_type",
+      this.get("item").attributes.package_type_id
+    );
+    this.set("pkgTypeType", pkgTypeType);
+  },
+
+  onLoad: function() {
+    this.set("loading", false);
+  },
+
+  is_cached: function(src) {
+    var image = new Image();
+    image.src = src;
+    return image.complete;
+  },
+
+  didInsertElement() {
+    var updateScreen = bind(this, this.onLoad);
+    this.$(".cl-item-image").on("load", updateScreen);
+  },
+
+  willDestroyElement() {
+    this.$(".cl-item-image").off("load");
+  }
+});
--- a/app/components/offers-block.js
+++ b/app/components/offers-block.js
@@ -1,6 +1,7 @@
 import { inject as service } from "@ember/service";
 import { getWithDefault } from "@ember/object";
 import Component from "@ember/component";
+import _ from "lodash";
 
 export default Component.extend({
   store: service(),
@@ -8,11 +9,32 @@ export default Component.extend({
 
   didReceiveAttrs() {
     this._super(...arguments);
+    console.log(this.get("record"));
     const location = this.get("store").peekRecord(
       "district",
       this.get("record").district_id
     );
+    this.getPackageTypesCounts(this.get("record").items);
     this.set("locationName", getWithDefault(location, "name", "N/A"));
     this.set("initialItems", this.get("record").items.slice(0, 4));
+  },
+
+  getPackageTypesCounts(items) {
+    const counts = _.reduce(
+      items,
+      (results, item) => {
+        const packageType = this.get("store").peekRecord(
+          "package_type",
+          item.attributes.package_type_id
+        );
+        const packageTypeName = getWithDefault(packageType, "name", "N/A");
+        const count = _.get(results, packageTypeName, 0);
+        _.set(results, packageTypeName, count + 1);
+        return results;
+      },
+      {}
+    );
+
+    this.set("typesCounts", counts);
   }
 });
--- a/app/components/offers-block.js
+++ b/app/components/offers-block.js
@@ -1,0 +1,18 @@
+import { inject as service } from "@ember/service";
+import { getWithDefault } from "@ember/object";
+import Component from "@ember/component";
+
+export default Component.extend({
+  store: service(),
+  locationName: "",
+
+  didReceiveAttrs() {
+    this._super(...arguments);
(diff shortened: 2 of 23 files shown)


app/components/item-thumbnail-view.js
```diff
@@ -0,0 +1,45 @@
...
+
+  changedSrc: function() {
+    if (!this.is_cached(this.get("src"))) {
+      this.set("loading", true);
+    }
+  }.observes("src"),
+
+  didReceiveAttrs() {
+    this._super(...arguments);
+    this.set("cloudinaryId", this.get("cloudinaryId"));
```
patrixr, line 18:
```
this.set("cloudinaryId", this.get("cloudinaryId"));
```

This doesn't do anything

app/components/item-thumbnail-view.js
```diff
@@ -0,0 +1,45 @@
...
+    this.set("src", this.generateUrl(500, 500, true));
+    const pkgTypeType = this.get("store").peekRecord(
+      "package_type",
+      this.get("item").attributes.package_type_id
+    );
+    this.set("pkgTypeType", pkgTypeType);
+  },
```
patrixr, line 25:
```javascript
  // no need for didReceiveAttrs, use computed properties
  src: Ember.computed('cloudinaryId', function () {
     return this.generateUrl(500, 500, true);
   }),
  
  pkgType: Ember.computed('packageTypeId', function () {
    // - I renamed  pkgTypeType into pkgType (didn't make any sense)
    // - you can set the packageTypeId one the the component instead of setting the item
    //  otherwise this component is not generic since it won't work with normal models (since you do `.attributes`)
    return this.get("store").peekRecord(
      "package_type",
      this.get("packageTypeId")
    );
  }),
```

app/components/item-thumbnail-view.js
```diff
@@ -0,0 +1,45 @@
...
+
+  onLoad: function() {
+    this.set("loading", false);
+  },
+
+  is_cached: function(src) {
```
patrixr, line 31:
```suggestion
  isCached: function(src) {
```

app/components/offers-block.js
```diff
@@ -0,0 +1,40 @@
...
+    const location = this.get("store").peekRecord(
+      "district",
+      this.get("record").district_id
+    );
+    this.getPackageTypesCounts(this.get("record").items);
+    this.set("locationName", getWithDefault(location, "name", "N/A"));
+    this.set("initialItems", this.get("record").items.slice(0, 4));
+  },
+
+  getPackageTypesCounts(items) {
```
patrixr, line 22:
```suggestion
  computePackageTypesCounts(items) {
```

Functions starting with "get" (`getXXX`) usually mean that they return something. This one doesn't return anything, it just sets stuff. Thus my renaming

app/components/offers-block.js
```diff
@@ -0,0 +1,40 @@
...
+    this._super(...arguments);
+    console.log(this.get("record"));
+    const location = this.get("store").peekRecord(
+      "district",
+      this.get("record").district_id
+    );
+    this.getPackageTypesCounts(this.get("record").items);
+    this.set("locationName", getWithDefault(location, "name", "N/A"));
+    this.set("initialItems", this.get("record").items.slice(0, 4));
+  },
```
patrixr, line 20: By doing `this.get("record").district_id` and `this.get("record").items`, you're making this component non-generic in the sense that it can't be used with any other data than exactly the one you're fetching. Not a very re-usable component

app/controllers/offers/index.js
```diff
@@ -0,0 +1,37 @@
...
+  showSearch: false,
+
+  selectedSort: computed({
+    get() {
+      return this.get("sortOptions.firstObject.id");
+    },
+    set(key, value) {
+      return value.id || value;
+    }
+  }),
```
patrixr, line 15: Don't think that's a good way to use getters and setters, note that the get doesn't change based on what the setter does. And the setter doesn't set anything

**Option 1**

```suggestion
  __selectedSort: null,
  selectedSort: computed({
    get() {
      return this.get('_selectedSort') || this.get("sortOptions.firstObject.id");
    },
    set(key, value) {
      this.set('_selectedSort', value);
      return value.id || value;
    }
  }),
```

**Option 2**

Don't use a computed property at all.

```suggestion
  init() {
    this.set('selectedSort', this.get("sortOptions.firstObject.id")); // set the default value and let the user modify it later
  }
```

app/controllers/offers/messages.js
```diff
@@ -0,0 +1,68 @@
...
+        })
+        .catch(error => {
+          this.store.unloadRecord(message);
+          throw error;
+        });
+    }
+  },
+
+  autoScroll() {
+    window.scrollTo(0, document.body.scrollHeight);
```
patrixr, line 27: as a general rule you want to avoid Jquery and DOM related code in controllers

app/controllers/offers/messages.js
```diff
@@ -0,0 +1,68 @@
...
+    function() {
+      return this.get("allMessages")
+        .filterBy("messageableType", "Offer")
+        .filterBy("messageableId", this.get("model.id"));
+    }
+  ),
+
+  actions: {
+    sendMessage() {
+      $("textarea").trigger("blur");
```
patrixr, line 47: Same comment as above

app/routes/offers/index.js
```diff
@@ -0,0 +1,41 @@
+import PublicRoute from "../browse_pages";
+import { inject as service } from "@ember/service";
+import _ from "lodash";
+
+export default PublicRoute.extend({
+  offerService: service(),
+
+  normalizeResponse(response) {
+    const normalizedResponse = _.reduce(
```
patrixr, line 9: probably better to use the term "packages" instead of "items"

app/routes/offers/index.js
```diff
@@ -0,0 +1,41 @@
...
+      response.data,
+      (results, offer) => {
+        const offerObj = {
+          ...offer.attributes
+        };
+        const items = _.filter(
+          response.included,
+          pkg => pkg.attributes.offer_id == offer.id
+        );
+        items.map(item => {
```
patrixr, line 19:
```suggestion
        items.forEach(item => {
```

app/router.js
```diff
@@ -13,6 +13,14 @@ Router.map(function() {
   this.route("package_category", {
     path: "/category/:id"
   });
+  this.route("offers", function() {
+    this.route("detail", {
+      path: "/:offer_id"
```
patrixr, line 18: Maybe rename this to `/:offer_public_uid` to be clear.

patrixr (approved):
@abulsayyad123 

I've added quite a few comments here and there, but it's mostly best practice stuff.
I'm approving this for merge, and you can improve the code in the next PRs (we're short on time sadly)

PR #620, opened 2021-02-12
GCW 3490 enhance shared offer feature

shreyas098:
### Ticket Link: 

https://jira.crossroads.org.hk/browse/GCW-3490

### What does this PR do?

FEATURE: Enhancing shareable feature in browse and removing ember scroll-to-bottom


app/components/item-thumbnail-view.js
```diff
@@ -1,45 +1,101 @@
...
-export default Component.extend(cloudinaryImage, {
-  loading: true,
+export default ImagePreview.extend(cloudinaryImage, {
   store: service(),
 
-  changedSrc: function() {
-    if (!this.is_cached(this.get("src"))) {
-      this.set("loading", true);
+  initializeLightgallery() {
+    var gallery = $(`#lightGallery${this.get("item.id")}`).data("lightGallery");
```
bharat619, line 11:
```suggestion
    const gallery = $(`#lightGallery${this.get("item.id")}`).data("lightGallery");
```

app/components/item-thumbnail-view.js
```diff
@@ -1,45 +1,101 @@
...
-  }.observes("src"),
 
-  didReceiveAttrs() {
-    this._super(...arguments);
-    this.set("cloudinaryId", this.get("cloudinaryId"));
-    this.set("src", this.generateUrl(500, 500, true));
-    const pkgTypeType = this.get("store").peekRecord(
-      "package_type",
-      this.get("item").attributes.package_type_id
+    var lightGalleryObj = $(`#lightGallery${this.get("item.id")}`).lightGallery(
```
bharat619, line 16: make use of const here as well

shreyas098: sure

app/components/item-thumbnail-view.js
```diff
@@ -1,45 +1,101 @@
...
+        selector: ".imageZoom"
+      }
     );
-    this.set("pkgTypeType", pkgTypeType);
+    this.set("lightGalleryObj", lightGalleryObj);
   },
 
-  onLoad: function() {
-    this.set("loading", false);
+  safeGet(model, id) {
```
bharat619, line 34: This is nice and handy function. Might be useful in many places

You can create a mixin for it and making it more configurable

```javascript
export default Ember.Mixin.create({
   safeGet: function(model, id, attr,defaultValue="N/A") {
       if(!attr) {
         throw new Error('Attribute cannot be empty');
       }
       
       if(!defaultValue) {
          throw new Error('Default value cannot be empty')
       }
       
       const record = this.get("store").peekRecord(model, id);
       return getWithDefault(record, "name", attr, defaultValue);
   }
})
```

app/components/item-thumbnail-view.js
```diff
@@ -1,45 +1,101 @@
...
+    function() {
+      return this.safeGet(
+        "donor_condition",
+        this.get("item").attributes.donor_condition_id
+      );
+    }
+  ),
+
+  dimension: computed("item", "item.id", function() {
+    let length = _.get(this.get("item"), "attributes.length");
```
bharat619, line 77: for `length` you have used lodash get, and for `width` and `height` you have used the normal `get`
want to make that consistent?

app/components/offers-block.js
```diff
@@ -19,9 +20,26 @@ function safeGet(obj, field, defaultValue) {
...
-export default Component.extend({
+export default Component.extend(cloudinaryImage, {
   store: service(),
 
+  didReceiveAttrs() {
+    this._super(...arguments);
+
+    this.get("items").forEach(item => {
+      const packageTypeName = this.typeOf(item);
+      _.set(item, "packageTypeName", packageTypeName);
```
bharat619, line 31: just to keep consistency, instead of using lodash here to set item
better to go with `item.set('packageTypeName', packageTypeName)`

shreyas098: item is a js object and not ember object so default way of set and get does not work here hence i used lodash

app/components/item-thumbnail-view.js
shreyas098: Thanks...i have added the mixin for it

abulsayyad123 (approved):

PR #624, opened 2021-03-12
GCW-3543-Glimmer issue fix

shreyas098:
On login,  user data was not loaded in the local storage hence cloud cart could not sync and the cart seems empty. When the user tries to add a package already present in the cloud cart (requested packages) the glimmer issue occurs

### DIFF
--- a/app/services/session.js
+++ b/app/services/session.js
@@ -40,7 +40,7 @@ export default ApiService.extend({
       return null;
     }
     return this.get("store").peekRecord("user", this.get("currentUserId"));
-  }).volatile(),
+  }),
 
   clear() {
     this.set("currentUserId", null);


bharat619 (approved):

swatijadhav (approved):

patrixr (approved):
@shreyas098 I have my doubts as whether this is a real fix, looks more like a workaround. Having a volatile ember prop should not be the cause for this error.

I'm approving given the urgency, but could you please test the following before releasing:

- Log in with User A
- Look at User A's dashboard (orders, name, chats, etc)
- Log out of session (do NOT refresh)
- Log in with User B
- Check to make sure User B's information is showing, and nothing from User A

abulsayyad123 (approved):

PR #648, opened 2021-08-05
GCW-2995 Fixed issue multiple initialization of websocket  

swatijadhav:
### Ticket Link: 
https://jira.crossroads.org.hk/browse/GCW-2995

### What does this PR do?
Issue: The application controller initializer was called multiple times and hence websocket.connect was called multiple times.
Since application controller is inherited in multiple other controllers, initializer block was called everytime.

Fixed issue by separating controllers into different parent controller.


app/controllers/application.js
```diff
@@ -1,214 +1,12 @@
...
-  ios_app_id: config.APP.APPLE_APP_ID,
-  appTitle: config.APP.TITLE,
-  bannerImage: config.APP.BANNER_IMAGE,
-  bannerReopenDays: config.BANNER_REOPEN_DAYS,
+export default applicationController.extend({
+  init() {
+    this._super(...arguments);
 
-  initializer: on("init", function() {
-    this.get("subscription").wire();
```
patrixr, line 36: @swatijadhav Is the point of all of this to prevent `wire()` from being called for all instances of application controller ?

If so this probably not the right way to fix this, you can simply add the `if (this.isSubscribed) return` statement inside the subscription service's wire method (services are instantiated only once and shared everywhere)

Also, I don't think there's a good reason for all those controllers to inherit from `applicationController`, we can probably remove that. The application controller can be injected into other controllers if they need to call some methods against it

Cutting the controller in half with `application.js` and `application_root.js` is too hacky

patrixr (changes requested):

app/services/subscription.js
```diff
@@ -118,6 +120,11 @@ export default Service.extend(Evented, {
   },
 
   wire() {
+    if (this.isSubscribed) {
+      return;
+    }
+
+    this.isSubscribed = true;
```
patrixr, line 127: @swatijadhav Looks good, just don't forget to set it back to false in the `unwire()` method

shreyas098 (approved):

